PR #2, opened 2021-11-12
Fix comment of seed PR

AbcSxyZ:
Fix last comment from PR https://github.com/dogecoin/dogecoin/pull/2337 about base image + workdir variable

### DIFF
--- a/1.14.4/x86_64-bionic/Dockerfile
+++ b/1.14.4/x86_64-bionic/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:focal
 
 # To improve : static hash make dynamic build of versions impossible.
 ARG VERSION=1.14.4
@@ -37,7 +37,7 @@ RUN cd /tmp && \
 
 COPY docker-entrypoint.py /usr/local/bin/docker-entrypoint
 
-WORKDIR /${USER}
+WORKDIR ${HOME}
 
 # P2P network (mainnet, testnet & regnet respectively)
 EXPOSE 22556 44556 18444
--- a/1.14.4/x86_64-bionic/docker-entrypoint.py
+++ b/1.14.4/x86_64-bionic/docker-entrypoint.py
@@ -6,6 +6,7 @@ import pwd
 import shutil
 import sys
 import subprocess
+import re
 
 def container_executable():
     """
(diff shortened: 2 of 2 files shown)


AbcSxyZ:
Error on pushed commits

1.14.4/x86_64-bionic/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:focal
```
patricklodder, line 1: Need to either rename the folder to `x86_64-focal` or keep this bionic for now. I'd recommend the latter, then we just create a focal (or maybe better, because smaller AND newer: `debian:bullseye`) image for 1.14.5.

AbcSxyZ: I didn't read carefully the folder name, I missed the version.

I would say we shouldn't have subfolder for distro version nor plateform. Virtualization should deal with it. I've to try the behavior on a raspberry if a non-arm executable will work inside the container.

Take a look at [mysql Dockerfile](https://github.com/docker-library/mysql/tree/15fe7357b165ee8aaa3ce165386f910a53a75087), it's by version and not by platform, they are only using `debian:buster-slim` to provide image for versions 5.6, 5.7 & 8.0.

AbcSxyZ: Possibly the repository architecture will be a single Dockerfile by version. And it would be definitively easier to maintain.

patricklodder: > I would say we shouldn't have subfolder for distro version nor plateform.

Where would I put the Dockerfiles for alpine 3.13 and 3.14 images? Dogecoin Core [got patched for musl compatibility](https://github.com/dogecoin/dogecoin/pull/2235) and we can support it...

> Virtualization should deal with it

We have multi-platform binaries, but you're saying is it's better to run x86 binaries on ARM through emulation?

> they are only using `debian:buster-slim`

I don't think we should exclude for example Alpine users just because that's a design choice that was made for mysql. Other official packages made different choices, eg: [postgres](https://github.com/docker-library/postgres/tree/master/10), [redis](https://github.com/docker-library/redis), [extreme example: nodejs](https://github.com/nodejs/docker-node), and so on.

Even if we were to choose to just support 1 distro (eg: debian) then there's still the issue that the Dogecoin Core release cycles are not at all in sync with distro release cycles. Since the goal **must** be to make production images (otherwise we may as well just keep all this in user repos and not care) we will have to think a bit about upgrade policies. You don't want to force updates cross distro release but instead enable custom security policies, the former will just lead to no one using the images we are proposing here at some point. I think that we'll have to maintain, say, `dogecoin:1.14.6-bullseye-slim` and `dogecoin:1.14.6-bookworm-slim` concurrently in the most minimal of supported cases.

PR #3, opened 2021-11-12
Use regex instead of man command to extract options

AbcSxyZ:
**Current Dockerfile can break with the following issue with minimized image like focal.**

To use environment variables to configure the container with `docker run`, the `docker-entrypoint.py` was using `man` command through a subprocess to get all existing option of an executable.

On minimized version of Ubuntu image, `man` command is disabled with the following message:
```
This system has been minimized by removing packages and content that are
not required on a system that users do not log into.

To restore this content, including manpages, you can run the 'unminimize'
```

Use python regex on raw man file to avoid the issue and retrieve all options.

## Potential Error

Using a minimized image could raise the following:
```
Traceback (most recent call last):
  File "/usr/local/bin/docker-entrypoint", line 124, in <module>
    EXECUTABLE_ARGS = convert_env()
  File "/usr/local/bin/docker-entrypoint", line 79, in convert_env
    dogecoind_options = executable_options()
  File "/usr/local/bin/docker-entrypoint", line 32, in executable_options
    raw_options = subprocess.check_output(
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'bash -c "man dogecoind | grep '^ *-'"' returned non-zero exit status 1.
```

### DIFF
--- a/1.14.4/x86_64-bionic/docker-entrypoint.py
+++ b/1.14.4/x86_64-bionic/docker-entrypoint.py
@@ -6,6 +6,7 @@ import pwd
 import shutil
 import sys
 import subprocess
+import re
 
 def container_executable():
     """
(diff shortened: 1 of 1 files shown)


patricklodder (approved):
Tested ACK.

Found that -e doesn't support ALL options (e.g. those from -help-debug, as those don't get added to the `manpage`) but since these work by just adding them as an argument to command space of `docker run`, I think we're good for now.

AbcSxyZ:
> Found that -e doesn't support ALL options (e.g. those from -help-debug, as those don't get added to the manpage)

I will take a look, I was able to grab it with the regex, I had to care about it specifically. It should be a quick thing to solve.

patricklodder:
> I will take a look, I was able to grab it with the regex, I had to care about it specifically. It should be a quick thing to solve.

Yeah we can pick this up separately, it's not a blocker because for now it works with `docker run` and those are mostly used for debugging / special circumstances. The only ones I can think of outside of debugging are, in sort-of order of importance: 

- `-sendfreetransactions` for allowing free tx right now
- `-checkblocks` for detecting HW problems (or in case of k8s, persistence layer issues - not uncommon)
- `-rpcservertimeout` for nodes that get pounded with slow RPC requests that have a cs_main lock, eg `getchaintips`
- `-rpcnamecoinapi` for miners
- `-logtimemicros` for nodes running debug logs
- `-limitfreerelay` in the future when we bring back free space for utxo consolidation

AbcSxyZ:
How were you passing the environment variable for `-help-debug` ? Did you use `docker run -e HELP_DEBUG=` exactly, with the equal ?

I verified and the argument is working, it's used with `execve` parameters.

patricklodder:
no that is ok. try `LOGTIMEMICROS=` or `RPCNAMECOINAPI=1`

AbcSxyZ:
There is no mention of `-logtimemicros` & `-rpcnamecoinapi` in the `dogecoind` man (1.14.4 & 1.14.5). These options do not exist for the program.

patricklodder:
```
docker run -e HELP= -e HELP_DEBUG= | grep -logtimemicros
```

edit: forgot we're testing 1.14.4 so there is no `-rpcnamecoinapi` because I added it in .5, but there is `-logtimemicros`

PR #4, opened 2021-11-12
Replace WORKDIR variable by HOME for readability

AbcSxyZ:
Fix last easy fix from comments of https://github.com/dogecoin/dogecoin/pull/2337.


patricklodder (approved):
tested ACK.

Issue #7, opened 2021-11-14
Repository Architecture : image base distribution and architecture


AbcSxyZ:
*Following comments from PR https://github.com/dogecoin/docker/pull/2#discussion_r748547687 about used distribution & architecture for the Dockerfile.*

I suggest speaking about architecture of the repository to see what would be possible to do to deal with versions, cross-platform portability & distributions. To clarify and move in a common direction.

### Cross-platform portability

I wasn't sure, but Docker do not manage hardware abstraction, because it uses the kernel, setting up a predefined software layer.

ARM & other specific architecture would need some tricks to be handled. Docker & Docker Hub are providing two ways to manage cross-platform images, the simplest way is using `docker buildx`, the alternative is to create manually a manifest from multiple published image/tag using `docker manifest`. See [docker blog](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/).

**`docker buildx` seems to not work,** x86_64 executable is not able to launch on ARM even with qemu emulation, creating an `exec format error`, not sure why.

Alternatively, building a root manifest from multiple tag is possible, resulting in a single version endpoint to call for all architectures. Multiple tag will be pushed for each architecture, using with the appropriate executable, and then everything is grouped under the same tag.
For now, it's what I'm having.

**You can find created images for both method on my [Docker Hub](https://hub.docker.com/r/abcsxyz/dogecoin/tags).**

So I've no certitude about the good path, but maybe `docker manifest` is something to explore. For now, I don't know if we have to create a folder for each architecture and how to publish all Dockerfiles. What is your current opinion ?

### Providing multiple distribution

I'm not clear about the benefit of providing multiple distros. I know about alpine for the size, @patricklodder was speaking about upgrades : 
> Even if we were to choose to just support 1 distro (eg: debian) then there's still the issue that the Dogecoin Core release cycles are not at all in sync with distro release cycles. Since the goal must be to make production images (otherwise we may as well just keep all this in user repos and not care) we will have to think a bit about upgrade policies. You don't want to force updates cross distro release but instead enable custom security policies, the former will just lead to no one using the images we are proposing here at some point.
I think that we'll have to maintain, say, dogecoin:1.14.6-bullseye-slim and dogecoin:1.14.6-bookworm-slim concurrently in the most minimal of supported cases. 

[...]

PR #8, opened 2021-11-14
1.14.5 node for debian:bookworm-slim

xanimo:
For after gitian verification moves from draft to publish and gets merged in main.


AbcSxyZ:
I suggest finding how we will organize the repository as raised by #7, let's use a single Dockerfile to work, and when we have an idea of the direction it will be easy to do this kind of change.

Let's find how we deal with architectures and distro :)

PR #10, opened 2021-11-16
Keep environment for execve + pop variables from environment

AbcSxyZ:
I thought removing the environment could be a great idea for security, for executable arguments at least. I was doubting and think I was wrong, it may bring to some unexpected behavior if some environment variables are needed.

Actually, I don't even know if the core use environment variable, but probably x)

Just removing environment variables passed through `docker run` for the executable, and leave it as it is after that.


patricklodder (approved):
tested ACK, this works well.

PR #13, opened 2021-11-17
Add full documentation for image

AbcSxyZ:
Create a draft for the image documentation. Can be the base to use for the readme available both from GitHub and from DockerHub.

You can have an overview of the result on the following [Docker Hub repository](https://hub.docker.com/r/abcsxyz/dogecoin).

*Inspired by documentation of official [Docker Hub images](https://github.com/docker-library/).*

**I want also to discuss (at least) the following line:**
```bash
docker exec (--user dogecoin ??) dogecoin-node dogecoin-cli help
```
When using `docker exec`, commands are launched as root. Idk if we specify to use a user ? We may specify to use the cli a different way, by creating another a new container for example.

**Edit:** Executable launched with `docker exec` are actually under `dogecoin` user rights, because I've setuid binaries. `docker exec` start as root and then switch to user.


AbcSxyZ:
I think it's something we can start to consider. I did some changes about `docker exec`, executable are setuid are running under `dogecoin` user in any case.

I created a separate `IMAGE_README.md` for the image documentation. I think this file could be used to explain image on Docker Hub.

This way we could have 2 README, with the main README of this repository explaining how to build manually images, run tests, develop & contribute, tools doc, etc, focusing more on image development. With the main README which still link to `IMAGE_README.md` to explain usage for those who use this path.

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+
+- [`1.14.5`,`latest`](#)
+
+## Quick reference
+
+- **Maintained by the [Dogecoin community](https://github.com/dogecoin/docker/)**
+- **Docker Hub images:** https://hub.docker.com/r/dogecoin/dogecoin
+- **Dogecoin Core repository:** https://github.com/dogecoin/dogecoin
+- **Where to file issues:** https://github.com/dogecoin/docker/issues
+- **Supported architectures:** [...]
```
patricklodder: This can be filled out

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+## Supported tags and respective `Dockerfile` links
+
+- [`1.14.5`,`latest`](#)
+
+## Quick reference
+
+- **Maintained by the [Dogecoin community](https://github.com/dogecoin/docker/)**
+- **Docker Hub images:** https://hub.docker.com/r/dogecoin/dogecoin
+- **Dogecoin Core repository:** https://github.com/dogecoin/dogecoin
+- **Where to file issues:** https://github.com/dogecoin/docker/issues
```
patricklodder:
```suggestion
**Issue tracker:** https://github.com/dogecoin/docker/issues
```

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+    ports:
+      - 22555:22555
+      - 22556:22556
+```
+Then, run `docker compose up`.
+
+## License
+
+Docker repository for Dogecoin Core is released under [MIT license](https://github.com/dogecoin/docker/blob/main/LICENSE).
+Feel free to use, modify & distribute for commercial and non-commercial purpose.
```
patricklodder, line 129: Not sure about this. It could be fine but am not a lawyer.

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+```bash
+# Syntax
+docker run  [docker-options] dogecoin/dogecoin [executable] [executable-options]
+```
+
++ `docker-options` : Set environment variables, ports, volumes and other docker settings.  
++ `executable` : Choose between `dogecoind`, `dogecoin-cli`, `dogecoin-tx` or `dogecoin-qt`. Default to `dogecoind`.  
++ `executable-options` : Pass options directly to the executable.
+
+There is three ways to configure the Dogecoin node. By using environment variables, by passing arguments to the executable or by providing `dogecoin.conf` in a volume.
```
patricklodder:
```suggestion
There are three ways to configure the Dogecoin node:

1. by using environment variables,
2. by passing arguments to the executable,
3. by providing a `dogecoin.conf` in a volume.
```

I think we should also document how these interact with each other, but currently, arguments override env, and I am not convinced that that's correct, so until then, just listing these will have to do.

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+### Configure with executable arguments
+
+`docker run` arguments are passed directly to the executable.
+```bash
+docker run dogecoin/dogecoin -paytxfee=0.01 -testnet
+```
+
+### Configure with environment variables
+
+All options for each executable can be defined with environment variables. Variables will be converted to executable arguments.
```
patricklodder: Suggest keeping this simpler:

```suggestion
All options, for each executable, can be defined with environment variables.
```

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+
+```bash
+docker run -e PAYTXFEE=0.01 -e TESTNET=1 dogecoin/dogecoin
+```
+Environment variables represent executable arguments converted in upper case, and `-` to `_`.  
+For example, `-help-debug` will be `HELP_DEBUG`.
```
patricklodder:
```suggestion
Environment variables represent executable arguments converted into upper case, without leading hyphen (`-`). 
Any hyphens inside an argument name must be converted to underscores (from `-`, to `_`.) 
Note that boolean arguments require the value `1` assigned to the variable.

For example, the `-help-debug` argument becomes `HELP_DEBUG=1` as environment variable.
```

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+Put this file in the directory mounted to `/dogecoin/.dogecoin`, in `datadir` for this example :
+```bash
+docker run -v $(pwd)/datadir:/dogecoin/.dogecoin dogecoin/dogecoin
+```
+
+### Docker Compose
+
+Example for `docker-compose.yml` :
```
patricklodder, line 107: May want to add a volume assignment to this?

IMAGE_README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+version: '3.1'
+
+services:
+  dogecoin:
+    image: dogecoin/dogecoin
+    command: --default-authentication-plugin=mysql_native_password
```
patricklodder: what does this do?

patricklodder (changes requested):
Some minor comments inline. I would suggest we get this ready, but not publish it before we are capable of publishing the images described here.

IMAGE_README.md
AbcSxyZ: Copy/paste error. Fixed

IMAGE_README.md
AbcSxyZ: done

IMAGE_README.md
AbcSxyZ: About what ? That's the permission under MIT, take a look at the permission recap github provide on [LICENSE](https://github.com/dogecoin/docker/blob/main/LICENSE) file.

People could sell it if they want, and of course use/modify/share.

IMAGE_README.md
AbcSxyZ: fixed

IMAGE_README.md
AbcSxyZ: Yeah, we can raise an issue to speak about this behavior.

AbcSxyZ:
In my opinion, it would be interesting to add this documentation when it's ready. For now, it's stated in big that the project is under development (we can add it at the top of this doc too), if someone come here he could be aware of this and still try to play with the image. Maybe some potential contributor could come and have something a bit documented to try.

If someone follows the doc, he will be quickly stopped by not being able to download the image from `dogecoin/dogecoin`. Adventurous people could slowly start to tests it, but it's still not easily available with Docker Hub nor Dogecoin Core repo, it limits widely the scope of people who will find it and think it's ready by mistake.

For me we should even add a link into the main repo, saying it's not finished, but a beginning of a Dockerfile is testable (but definitively unstable, available for development only). It would facilitate people participation.

PR #14, opened 2021-11-20
multi-architecture implementation

### LINKED ISSUE #17: feat: Support non-complicated x86_64 builds.
Post-merge feature after #14:

If we can find a way to give the `TARGETPLATFORM` build argument a default value of `linux/amd64`, we can make the build process simpler.

_From: https://github.com/dogecoin/docker/pull/14#r754617790_

AbcSxyZ:
fix https://github.com/dogecoin/docker/issues/17.

Beginning of an implementation for multi-architecture support using `buildx`. Remove security check, to rely on #6.

I'm able to build images for those for architecture : `amd64`, `arm64`, `armv7` & `i386`.
[Here](https://hub.docker.com/r/abcsxyz/dogecoin/tags) for Docker hub example.

It may be possible to use `dpkg` or `TARGETPLATFORM` to find architectures. It seems to me that `TARGETPLATFORM` is easier to use with `buildx`, but won't work if used directly with `docker build`.
`dpkg` would need few more operations (install of `dpkg-dev`, set manually variables before a dirtier mapping), but would work with `docker build`. Docker official images are using `dpkg`.

This version is cleaner for pushing to Docker Hub, but may be interesting to let the user use the image manually using `dpkg`. Any opinion ?

*It's using Github Action with configuration defined in #15 to publish image on Docker Hub*

### DIFF
--- a/1.14.4/x86_64-bionic/Dockerfile
+++ b/1.14.4/x86_64-bionic/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:bionic
 
 # To improve : static hash make dynamic build of versions impossible.
 ARG VERSION=1.14.4
-ARG VERSION_HASH=6266235abe4bcbd41ea57bdf42f11ef89aa69f0386e8c8846d5228af69e7fa13
+ARG TARGETPLATFORM
 
 ENV USER=dogecoin
 ENV DATADIR=/${USER}/.dogecoin
@@ -19,23 +19,28 @@ RUN apt update && apt install -y \
     wget \
     && rm -rf /var/lib/apt/lists/*
 
-# Download Dogecoin Core from github releases.
+# Download Dogecoin Core from github releases,
+# manage binary architecture using buildx & TARGETPLATFORM.
+WORKDIR /tmp
+
+RUN set -ex && \
+    if [ "${TARGETPLATFORM}" = "linux/amd64" ]; then TARGETPLATFORM=x86_64-linux-gnu; fi \
+    && if [ "${TARGETPLATFORM}" = "linux/arm64" ]; then TARGETPLATFORM=aarch64-linux-gnu; fi \
+    && if [ "${TARGETPLATFORM}" = "linux/arm/v7" ]; then TARGETPLATFORM=arm-linux-gnueabihf; fi \
+    && if [ "${TARGETPLATFORM}" = "linux/386" ]; then TARGETPLATFORM=i686-pc-linux-gnu; fi \
+    && wget https://github.com/dogecoin/dogecoin/releases/download/v${VERSION}/dogecoin-${VERSION}-${TARGETPLATFORM}.tar.gz
+
 # Move downloaded binaries and man pages in the container system.
-# Setuid on binaries with $USER rights, to limit root usage.
-#
-# Security: more secure way than check hash for download,
-# see https://github.com/docker-library/official-images#security
-RUN cd /tmp && \
-    wget https://github.com/dogecoin/dogecoin/releases/download/v${VERSION}/dogecoin-${VERSION}-x86_64-linux-gnu.tar.gz && \
-    echo "${VERSION_HASH} dogecoin-${VERSION}-x86_64-linux-gnu.tar.gz" | sha256sum -c && \
-    tar -xvf dogecoin-${VERSION}-x86_64-linux-gnu.tar.gz --strip-components=1 && \
+# Setuid on binaries with $USER rights, to prevent
+# root right with `docker exec`.
+RUN tar -xvf dogecoin-${VERSION}-*.tar.gz --strip-components=1 && \
     cp share/man/man1/*.1 /usr/share/man/man1 && \
     cp bin/dogecoin* /usr/local/bin && \
     chown ${USER}:${USER} /usr/local/bin/dogecoin* && \
     chmod 4555 /usr/local/bin/dogecoin* && \
     rm -rf /tmp/*
 
-COPY docker-entrypoint.py /usr/local/bin/docker-entrypoint
+COPY --chmod=555 docker-entrypoint.py /usr/local/bin/docker-entrypoint
 
 WORKDIR ${HOME}
 
--- a/1.14.4/x86_64-bionic/Dockerfile
+++ b/1.14.4/x86_64-bionic/Dockerfile
@@ -2,7 +2,6 @@ FROM ubuntu:bionic
 
 # To improve : static hash make dynamic build of versions impossible.
 ARG VERSION=1.14.4
-ARG TARGETPLATFORM
 
 ENV USER=dogecoin
 ENV DATADIR=/${USER}/.dogecoin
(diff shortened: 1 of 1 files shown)


1.14.4/x86_64-bionic/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:bionic
 
 # To improve : static hash make dynamic build of versions impossible.
 ARG VERSION=1.14.4
-ARG VERSION_HASH=6266235abe4bcbd41ea57bdf42f11ef89aa69f0386e8c8846d5228af69e7fa13
+ARG TARGETPLATFORM
```
patricklodder: Can we give this a default value of `linux/amd64`? Otherwise specifying this becomes a requirement for building.

1.14.4/x86_64-bionic/Dockerfile
```diff
@@ -19,23 +19,28 @@ RUN apt update && apt install -y \
...
+# root right with `docker exec`.
+RUN tar -xvf dogecoin-${VERSION}-*.tar.gz --strip-components=1 && \
     cp share/man/man1/*.1 /usr/share/man/man1 && \
     cp bin/dogecoin* /usr/local/bin && \
     chown ${USER}:${USER} /usr/local/bin/dogecoin* && \
     chmod 4555 /usr/local/bin/dogecoin* && \
     rm -rf /tmp/*
 
-COPY docker-entrypoint.py /usr/local/bin/docker-entrypoint
+COPY --chmod=500 docker-entrypoint.py /usr/local/bin/docker-entrypoint
```
patricklodder:
```
Step 12/17 : COPY --chmod=500 docker-entrypoint.py /usr/local/bin/docker-entrypoint
the --chmod option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
```

Are we sure we want that dependency?

patricklodder (commented):
Noted some things I noticed when testing this inline.

1.14.4/x86_64-bionic/Dockerfile
AbcSxyZ: If we set this value, the build won't be portable. That's what I commented in the PR comment. `TARGETPLATFORM` is appropriate to get a cleaner Dockerfile while using `buildx`. To have an image to build with `docker build` command, we may need to use `dpkg` to get the architecture.

1.14.4/x86_64-bionic/Dockerfile
AbcSxyZ: I didn't know `chmod` was specific to `buildx`. Depend on the previous question about `buildx` vs `dpkg`.

patricklodder: It requires buildkit, which is not enabled by default everywhere. I think we can use `RUN chmod 500` too?

1.14.4/x86_64-bionic/Dockerfile
patricklodder: So buildx will not `--build-arg`, but instead do magic that cannot override a default value? Ouch! I'll do some more research, there must be a way to make it work with both.

xanimo: "When you invoke a build, you can set the --platform flag to specify the target platform for the build output, (for example, linux/amd64, linux/arm64, or darwin/amd64)."

This is from working with buildx.

xanimo: Can confirm passing --platform populates ARG TARGETPLATFORM and is accessible with $TARGETPLATFORM

AbcSxyZ: > So buildx will not --build-arg, but instead do magic that cannot override a default value? 

Idk exactly how it's done, but probably it's using --build-arg. If you don't use ARG keyword, TARGETPLATFORM isn't available. The issue is hard-coding an architecture as default value, when the image is will be build on a different architecture.

@xanimo I guess the best scenario would be to not force using `buildx`, and let users create the image using `docker build`.

1.14.4/x86_64-bionic/Dockerfile
AbcSxyZ: Sure. I would just recommend to use chmod within `COPY` if the image is only deployed/buildable with `buildx`. Otherwise, in `RUN` statement of course.

patricklodder: Why do you recommend it? Reduction of build steps or is there more to it?

AbcSxyZ: > Reduction of build steps 

Yes

1.14.4/x86_64-bionic/Dockerfile
patricklodder: Let's leave this for now, I will propose a solution for doing simple `docker build .` in a separate PR.

1.14.4/x86_64-bionic/Dockerfile
patricklodder: Okay. Could you please change that then? Let's keep the extra build step until buildkit is the default build process everywhere (currently it's only default on docker-desktop.)

AbcSxyZ: Rn, because we are using `TARGETPLATFORM`, someone who will build this Dockerfile will need `buildx`.

In any case not a big change, but don't you want to do this edit when/if we're having a solution to enable `docker build` command ?

If we want to go with `docker build`, I suppose we should use `dpkg`. I think it's a matter of choice between having a more concise Dockerfile using `buildx` solution, where people will use this image in their own Dockerfile or with the cli and won't build it manually. Or we can get something more verbose, possibly a dirtier mapping, but easily buildable with `dpkg`.

I like this idea of using `docker build`, the solution may be not that bad, but it's the "simpler" path here.
Let's continue in https://github.com/dogecoin/docker/issues/17.

patricklodder: I can do `docker build --build-arg TARGETPLATFORM=linux/amd64 .` right now to build this but it will fail on this instruction if one doesn't use [buildkit](https://docs.docker.com/develop/develop-images/build_enhancements/) by default. 

The reason why I am asking to make the separate `RUN chmod 500 /usr/local/bin/docker-entrypoint` for the time being is to retain some form of backward compatibility for now, until buildkit is the only supported method. Since this does not affect the outcome - it only costs a trivial build step - this de-optimization helps portability until a later time that the build system is based on buildkit everywhere...

I have no way to build this without buildkit for this issue, whereas for the first issue I mentioned there is a workaround.

AbcSxyZ: Let's supersede this design with work from https://github.com/dogecoin/docker/issues/17 :)

AbcSxyZ:
Add implementation from issue https://github.com/dogecoin/docker/issues/17 for cross architecture using `dpkg --print-architecture` instead of `TARGETPLATFORM` prepopulated variable.

patricklodder (approved):
ACK. 

Note: This (`dpkg`) works for all debian-based distros, so we should have no issues for now.

PR #19, opened 2021-11-29
Enable arbitrary shell commands for entrypoint

### LINKED ISSUE #18: Add shell commands support in entrypoint
From some documentation about [Docker Official Images](https://github.com/docker-library/official-images#consistency):

> All official images should provide a consistent interface. A beginning user should be able to docker run official-image bash (or sh) without needing to learn about --entrypoint. It is also nice for advanced users to take advantage of entrypoint, so that they can docker run official-image --arg1 --arg2 without having to specify the binary to execute.

Need to enable execution of any kind of command in the entrypoint. Need to follow one process if used command is related to dogecoin, otherwise launch arbitrary commands.

AbcSxyZ:
*Fix https://github.com/dogecoin/docker/issues/18*

Enable to launch arbitrary shell commands using `docker run`, like `docker run -it [image] bash`.

Following recommendation from [official docker images](https://github.com/docker-library/official-images#consistency):

> All official images should provide a consistent interface. A beginning user should be able to docker run official-image bash (or sh) without needing to learn about --entrypoint. It is also nice for advanced users to take advantage of entrypoint, so that they can docker run official-image --arg1 --arg2 without having to specify the binary to execute.

### DIFF
--- a/1.14.4/x86_64-bionic/Dockerfile
+++ b/1.14.4/x86_64-bionic/Dockerfile
@@ -48,3 +48,4 @@ EXPOSE 22555 44555 18332
 VOLUME ["/dogecoin/.dogecoin"]
 
 ENTRYPOINT ["docker-entrypoint"]
+CMD ["dogecoind"]
--- a/1.14.4/x86_64-bionic/docker-entrypoint.py
+++ b/1.14.4/x86_64-bionic/docker-entrypoint.py
@@ -8,23 +8,26 @@ import sys
 import subprocess
 import re
 
-def container_executable():
-    """
-    Get the executable used in the docker command, or
-    use dogecoind by default.
-    """
-    dogecoin_executables = [
-                "dogecoind",
-                "dogecoin-qt",
-                "dogecoin-cli",
-                "dogecoin-tx",
-            ]
+CLI_EXECUTABLES = [
+        "dogecoind",
+        "dogecoin-qt",
+        "dogecoin-cli",
+        "dogecoin-tx",
+        ]
 
-    executable = "dogecoind"
-    if len(sys.argv) >= 2 and sys.argv[1] in dogecoin_executables:
-        executable = sys.argv.pop(1)
+def execute(exectuable, args):
+    """
+    Run container command with execve(2). Use manually execve
+    to run the process as same pid and avoid to fork a child.
+    """
+    executable_path = shutil.which(executable)
 
-    return executable
+    if executable_path is None:
+        exit(f"{sys.argv[0]}: {executable} not found.")
+
+    #Prepare execve args & launch container command
+    execve_args = [executable_path] + args
+    os.execve(executable_path, execve_args, os.environ)
 
 def characters_cleaner(raw_option):
     """
@@ -115,12 +118,8 @@ def run_executable(executable, executable_args):
     to manage a single process in a container & more predictive
     signal handling.
     """
-    #Prepare execve(2) arguments
     if executable in ["dogecoind", "dogecoin-qt"]:
         executable_args.append("-printtoconsole")
-
-    executable_path = shutil.which(executable)
-    execve_args = [executable_path] + executable_args
 
     #Switch process from root to user.
     #Equivalent to use gosu or su-exec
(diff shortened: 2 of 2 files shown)


AbcSxyZ:
**[Broken]**

Do not work when you're doing the following:
```
docker run -it [image] -testnet
```
Do not use `dogecoind` as default command.

AbcSxyZ:
Previous issue fixed. Following [mysql](https://github.com/docker-library/mysql/blob/37cf4047a0544046c9ecce8f2885d4f362a227bc/8.0/docker-entrypoint.sh#L363) strategy to handle `docker run` commands containing only options.

patricklodder (approved):
ACK.

Note: I had a thought about running the executable before `create_datadir()`, but I settled on this being the right place, because it allows to open a shell in case an error arises when the datadir gets created...

xanimo:
> **[Broken]**
> 
> Do not work when you're doing the following:
> 
> ```
> docker run -it [image] -testnet
> ```
the above command works on main if you add an additional `-` e.g.: `docker run -it [image] --testnet`
> 
> Do not use `dogecoind` as default command.
it still automatically executes if i run `docker run -it <image>` without options.

and if the purpose of this is to allow a beginner to get a bash shell why not use `docker exec -it <image> bash`?
also keep in mind that both commands will enter the container as root:
```
root@b6343fc79e8b:~# cd ../
root@b6343fc79e8b:/# ls -l
total 68
drwxr-xr-x   2 root root 4096 Sep 30 12:34 bin
drwxr-xr-x   2 root root 4096 Apr 24  2018 boot
drwxr-xr-x   5 root root  360 Dec  3 23:30 dev
drwxr-xr-x   1 root root 4096 Dec  3 23:30 dogecoin
drwxr-xr-x   1 root root 4096 Dec  3 23:30 etc
drwxr-xr-x   2 root root 4096 Apr 24  2018 home
drwxr-xr-x   1 root root 4096 May 23  2017 lib
drwxr-xr-x   2 root root 4096 Sep 30 12:33 lib64
drwxr-xr-x   2 root root 4096 Sep 30 12:32 media
drwxr-xr-x   2 root root 4096 Sep 30 12:32 mnt
drwxr-xr-x   2 root root 4096 Sep 30 12:32 opt
dr-xr-xr-x 457 root root    0 Dec  3 23:30 proc
drwx------   2 root root 4096 Sep 30 12:33 root
drwxr-xr-x   5 root root 4096 Sep 30 12:34 run
drwxr-xr-x   2 root root 4096 Sep 30 12:34 sbin
drwxr-xr-x   2 root root 4096 Sep 30 12:32 srv
dr-xr-xr-x  13 root root    0 Dec  3 23:30 sys
drwxrwxrwt   1 root root 4096 Dec  3 21:34 tmp
drwxr-xr-x   1 root root 4096 Sep 30 12:32 usr
drwxr-xr-x   1 root root 4096 Sep 30 12:33 var
```
which i've just tested deleting `rm -rf *` successfully so maybe we want to reconsider given the emphasis on security?

outside of that i've found a bug not solved by either main or this branch which is attempting to run as daemon:
`docker run -it <image> -daemon`, `docker run -it <image> --daemon`, `docker run -it <image> dogecoind -daemon`, etc. it starts to appear it works but then exits. forgive me if i'm missing something, am fine with the added functionality of simply running `docker run -it <image> bash` but am wondering if that's the only thing we're accomplishing outside the removing the extra `-` for args?

AbcSxyZ:
It's the expected behavior to enter as root, for example mysql is doing the same. Can't enter as user without specifying `--user` option. I do not see a way to achieve it, and it can also let user navigate in the container without trouble.

Nice spot for `-deamon`, not sure what is happening but possibly specific to this argument, not sure what's going on rn.

I'm getting the following error:
```
$ docker run -it dogecoin -daemon
2021-12-04 00:35:31 
2021-12-04 00:35:31 Dogecoin version v1.14.4
2021-12-04 00:35:31 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2021-12-04 00:35:31 Assuming ancestors of block 77e3f4a4bcb4a2c15e8015525e3d15b466f6c022f6ca82698f329edef7d9777e have valid signatures.
Dogecoin server starting
```

xanimo:
yep, that's the same output i was getting too. and ok regarding root. will
ACK in one sec and leave it up to you and pat to decide whether we should
solve the -daemon flag now or in a separate PR.

xanimo (approved):
ACK

PR #20, opened 2021-11-29
Work on 1.14.5 with debian:bullseye-slim

patricklodder:
Uses latest stable Debian slim and latest Dogecoin Core. 

As 1.14.4 has linux CVEs that got fixed in 1.14.5, and this has not yet been released, remove the entire 1.14.4 tree. 

We should not ever knowingly publish vulnerable software.


patricklodder:
(needs a rebase after #6)

AbcSxyZ:
What would be the good strategy to adopt ? That's the question.

Old softwares will have known vulnerabilities, it's a matter of time to have vulnerable versions published. Also, people may need previous version for some reasons, it can be something to deal with too.

I was looking at the strategy of `postgres` images, you can see their GitHub history for a specific version ([10 here](https://github.com/docker-library/postgres/commits/master?after=f5225fd4fb1dcc605df32819804f19284cea28b6+69&branch=master&path%5B%5D=10)) and corresponding tags on [Docker Hub](https://hub.docker.com/_/postgres?tab=tags&page=1&name=10).

They are:
- Updating GitHub folder of a version with the latest minor release.
- Leaving tags for each minor release on Docker Hub, having all versions available for download

The solution may be to publish all 1.14 release, having 1.14.5 pointing to `dogecoin:1.14.5`, `dogecoin:1.14`, `dogecoin:latest` ? Then the same for `1.21` ? Depend on the version strategy of Dogecoin Core. I'm not clear about what it is actually & how it works related to bitcoin.

patricklodder:
> Old softwares will have known vulnerabilities, it's a matter of time to have vulnerable versions published. 

Agree. Just, if we haven't published a vulnerable software yet and there is a patched alternative, we shouldn't.

> having 1.14.5 pointing to dogecoin:1.14.5, dogecoin:1.14, dogecoin:latest

This. And with `1.21` you'll have `dogecoin:1.21.0`, `dogecoin:1.21` and then that will be `dogecoin:latest`.

We need to define a QC gate for `:latest` and `:<major>`, which is going to be some more discussion.

However, what I'm saying is: **if we haven't released 1.21.0 and 1.21.1 solves a CVE, then we should not release 1.21.0 after the fact.**

AbcSxyZ:
>  if we haven't released 1.21.0 and 1.21.1 solves a CVE, then we should not release 1.21.0 after the fact.

All right.

AbcSxyZ (dismissed):

AbcSxyZ:
We have a functional Dockerfile :)

xanimo (dismissed):
ACK. was weird when building concurrently using buildx build --platform=, ubuntu uses linux/686 as shorthand for i386 whereas debian uses linux/386. the more you know heh..
https://hub.docker.com/repository/docker/xanimo/1.14.5-dogecoin

patricklodder:
Got conflicted with #19 - rebased, now it's okay again.

AbcSxyZ (approved):

xanimo (approved):
ACK
https://hub.docker.com/repository/registry-1.docker.io/xanimo/1.14.5-dogecoin/tags?page=1&ordering=last_updated

Issue #16, opened 2021-11-21
Script to generate Dockerfiles by version + Github Action integration


AbcSxyZ:
I'm starting to think about a script to generate a Dockerfile for each Dogecoin Core release, and to have a way to update existing Dockerfiles/images when a template is edited, by using GitHub Action (GA) to automate Docker Hub tasks.
The goal here is to draft the script & the GA process.

The repository architecture would be the following, as defined in [the related issue](https://github.com/dogecoin/docker/issues/7#issuecomment-970673781):
```
.
└── version
    ├── distro1
    │   └── Dockerfile
    └── distro2
        └── Dockerfile
└── version2
   ├── distro1
   ....
```
As I see it now, we can have 2 type of event : 
| GitHub change |  Docker Hub changes | 
|--------------------|----------------------------|
| push on release  | 1 version pushed, 1 image per distro |
|  update of templates | all version updated, for all distros  |

By reading GA documentation, I think it can be possible to achieve something like that :crossed_fingers:.

Should we update all Dockerfiles and images when a template is modified ? Docker Hub do not handle versioning for images, and changes can break related programs. Docker official images do not use a version strategy, but we may have change they don't.

I'm not sure how intensively GA should automate stuff. Do we try to do the publication of a new version with a workflow on Dogecoin Core called on releases ? Depend on the previous question, but do we automate update of templates (and Dockerfiles), GA propagating the change & committing automatically, or should do a PR ? Alternatively, do we let human run the script, creating locally new files & PR the result.

Any idea, expectation about the script/GA ? Any limitation to this current suggestion ?

AbcSxyZ:
Beginning of an implementation with PR https://github.com/dogecoin/docker/pull/40 about Dockerfile templatization.

georgeartem:
why would we want all versions? wouldn't it make better sense to have docker support for some version going forward like 1.14.7 or 1.14.9 or whatever - just make a choice on where the historical cutoff would be for unsupported versions

georgeartem:
When a new core version is released, selecting what is determined to be the most stable/performant asset as determined by the testing completed during the release cycle and pulling that into your docker ci would reduce the complexity proposed here significantly.

PR #11, opened 2021-11-16
Avoid using EXECUTABLE as global variable

AbcSxyZ:
Naming changes. It was using `EXECUTABLE` variable globally to access to the used executable. Avoid to do that, passing the variable as function argument, looks cleaner.

PR #23, opened 2021-12-03
qa: add build ci

patricklodder:
Adds a GH Actions CI for **testing pull requests** across all build targets.

Notes:
- This is NOT the same as the build & publish on push from #15. That PR takes care of publication on push, this PR takes care of testing PRs. The matrix parallelizes builds for each platform, which is not good for CD but extremely good for CI because it allows much easier debugging.
- Dynamic matrix generation was inspired by what [nodejs/docker-node](https://github.com/nodejs/docker-node/blob/main/genMatrix.js) does for multi-version, multi-platform builds. I wrote the script from scratch for it to fully fit our usecase.
- It will only test changed scope. So if there are 4 versions across 4 distros but we only change 2, we do not have to test 16*4 builds.
- Adds a flat file `PLATFORMS` to each version/platform pair (which is ugly) until we have settled on some sort of machine-readable manifest (I have a ton of thoughts about this, will create separate issue for it later.) Either way, different Dogecoin Core versions and different distros can support different platforms, so we need to document this somewhere...
-  We will want to add tests to this script. I have added a simple version check.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -1,0 +1,69 @@
+name: build-ci
+
+on:
+  pull_request:
+    paths:
+      - "**/Dockerfile"
+      - "**/docker-entrypoint.py"
+      - "**/PLATFORMS"
+      - "tools/genmatrix.js"
+      - ".github/workflows/build-ci.yml"
+
+jobs:
+  gen-matrix:
+    name: generate-matrix
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Get changed files
+        id: get-changed-files
+        uses: jitterbit/get-changed-files@v1
+        with:
+          format: 'json'
+
+      - name: Generate testing matrix
+        uses: actions/github-script@v4.1
+        id: generator
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const script = require(`${process.env.GITHUB_WORKSPACE}/tools/genmatrix.js`)
+            return script(process.env.GITHUB_WORKSPACE, ${{ steps.get-changed-files.outputs.all }});
+    outputs:
+      matrix: ${{ steps.generator.outputs.result }}
+
+  build:
+    if: ${{ fromJson(needs.gen-matrix.outputs.matrix) }}
+    needs: gen-matrix
+    name: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.gen-matrix.outputs.matrix) }}
+
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build image
+        uses: docker/build-push-action@v2
+        with:
+          builder: ${{ steps.buildx.outputs.name }}
+          push: false
+          load: true
+          platforms: ${{ matrix.platform }}
+          context: ./${{ matrix.version }}/${{ matrix.variant }}
+          file: ./${{ matrix.version }}/${{ matrix.variant }}/Dockerfile
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -4,8 +4,9 @@ on:
   pull_request:
     paths:
       - "**/Dockerfile"
-      - "**/docker-entrypoint.py"
+      - "**/entrypoint.py"
       - "**/PLATFORMS"
+      - "tests/"
       - "tools/genmatrix.js"
       - ".github/workflows/build-ci.yml"
 
@@ -38,6 +39,8 @@ jobs:
     if: ${{ fromJson(needs.gen-matrix.outputs.matrix) }}
     needs: gen-matrix
     name: build
+    env:
+      image_tag: local/ci:${{ github.run_id }}
     runs-on: ubuntu-latest
     timeout-minutes: 60
     strategy:
@@ -62,8 +65,16 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           push: false
           load: true
+          tags: ${{ env.image_tag }}
           platforms: ${{ matrix.platform }}
           context: ./${{ matrix.version }}/${{ matrix.variant }}
           file: ./${{ matrix.version }}/${{ matrix.variant }}/Dockerfile
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Run integration tests
+        run: |
+          python3 -m tests.integration_runner \
+                  --platform ${{ matrix.platform }} \
+                  --image ${{ env.image_tag }} \
+                  --version ${{ matrix.version }}
--- a/1.14.5/bullseye/PLATFORMS
+++ b/1.14.5/bullseye/PLATFORMS
@@ -1,0 +1,4 @@
+linux/amd64
+linux/arm64
+linux/arm/v7
+linux/386
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Dogecoin Core developers
+"""
+Test framework for end-to-end docker tests
+"""
+
+import subprocess
+import sys
+
+class DockerRunner:
+    """Run docker containers for testing"""
+
+    def __init__(self, platform, image, verbose):
+        """Sets platform and image for all tests ran with this instance"""
+        self.platform = platform
+        self.image = image
+        self.verbose = verbose
+
+    def construct_docker_command(self, envs, args):
+        """
+        Construct a docker command with env and args
+        """
+        command = ["docker", "run", "--platform", self.platform]
+
+        for env in envs:
+            command.append("-e")
+            command.append(env)
+
+        command.append(self.image)
+
+        for arg in args:
+            command.append(arg)
+
+        return command
+
+    def run_interactive_command(self, envs, args):
+        """
+        Run our target docker image with a list of
+        environment variables and a list of arguments
+        """
+        command = self.construct_docker_command(envs, args)
+
+        if self.verbose:
+            print(f"Running command: { ' '.join(command) }")
+
+        try:
+            output = subprocess.run(command, capture_output=True, check=True)
+        except subprocess.CalledProcessError as docker_err:
+            print(f"Error while running command: { ' '.join(command) }", file=sys.stderr)
+            print(docker_err, file=sys.stderr)
+            print(docker_err.stderr.decode("utf-8"), file=sys.stderr)
+            print(docker_err.stdout.decode("utf-8"), file=sys.stdout)
+
+            raise docker_err
+
+        return output
--- a/tests/integration/framework/test_runner.py
+++ b/tests/integration/framework/test_runner.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Dogecoin Core developers
+"""
+Base class to define and run Dogecoin Core Docker tests with
+"""
+
+import argparse
+import sys
+
+from .docker_runner import DockerRunner
+
+class TestConfigurationError(Exception):
+    """Raised when the test is configured inconsistently"""
+
+class TestRunner:
+    """Base class to define and run Dogecoin Core Docker tests with"""
+    def __init__(self):
+        """Make sure there is an options object"""
+        self.options = {}
+
+    def add_options(self, parser):
+        """Allow adding options in tests"""
+
+    def run_test(self):
+        """Actual test, must be implemented by the final class"""
+        raise NotImplementedError
+
+    def run_command(self, envs, args):
+        """Run a docker command with env and args"""
+        assert self.options.platform is not None
+        assert self.options.image is not None
+
+        runner = DockerRunner(self.options.platform,
+            self.options.image, self.options.verbose)
+
+        return runner.run_interactive_command(envs, args)
+
+    def main(self):
+        """main loop"""
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--platform", dest="platform", required=True,
+            help="The platform to use for testing, eg: 'linux/amd64'")
+        parser.add_argument("--image", dest="image", required=True,
+            help="The image or tag to execute tests against, eg: 'verywowimage'")
+        parser.add_argument("--verbose", dest="verbose", default=False, action="store_true",
+            help="Verbosely output actions taken and print docker logs, regardless of outcome")
+
+        self.add_options(parser)
+        self.options = parser.parse_args()
+
+        self.run_test()
+        print("Tests successful")
+        sys.exit(0)
--- a/tests/integration/version.py
+++ b/tests/integration/version.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Dogecoin Core developers
+"""
+Test the version installed to be the expected version
+"""
+
+import re
+
+from .framework.test_runner import TestRunner
+
+class VersionTest(TestRunner):
+    """Versions test"""
+
+    def __init__(self):
+        """Constructor"""
+        TestRunner.__init__(self)
+        self.version_expr = None
+
+    def add_options(self, parser):
+        """Add test-specific --version option"""
+        parser.add_argument("--version", dest="version", required=True,
+            help="The version that is expected to be installed, eg: '1.14.5'")
+
+    def run_test(self):
+        """Check the version of each executable"""
+
+        self.version_expr = re.compile(f".*{ self.options.version }.*")
+
+        # check dogecoind with only env
+        dogecoind = self.run_command(["VERSION=1"], [])
+        self.ensure_version_on_first_line(dogecoind.stdout)
+
+        # check dogecoin-cli
+        dogecoincli = self.run_command([], ["dogecoin-cli", "-?"])
+        self.ensure_version_on_first_line(dogecoincli.stdout)
+
+        # check dogecoin-tx
+        dogecointx = self.run_command([], ["dogecoin-tx", "-?"])
+        self.ensure_version_on_first_line(dogecointx.stdout)
+
+        # make sure that we find version errors
+        caught_error = False
+        try:
+            self.ensure_version_on_first_line("no version here".encode('utf-8'))
+        except AssertionError:
+            caught_error = True
+
+        if not caught_error:
+            raise AssertionError("Failed to catch a missing version")
+
+    def ensure_version_on_first_line(self, cmd_output):
+        """Assert that the version is contained in the first line of output string"""
+        first_line = cmd_output.decode("utf-8").split("\n")[0]
+
+        if re.match(self.version_expr, first_line) is None:
+            text = f"Could not find version { self.options.version } in { first_line }"
+            raise AssertionError(text)
+
+if __name__ == '__main__':
+    VersionTest().main()
--- a/tests/integration_runner.py
+++ b/tests/integration_runner.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Dogecoin Core developers
+"""
+Runs the integration tests
+"""
+
+import subprocess
+import sys
+
+from .integration.framework.test_runner import TestRunner
+
+def print_test_output(test_name, stdout, stderr=None):
+    """prints output from a test, including from stderr if provided"""
+    print("\n")
+    print(test_name)
+    print("----------------------")
+
+    if stderr is not None:
+        print(stderr.decode("utf-8"), file=sys.stderr)
+
+    print(stdout.decode("utf-8"))
+
+class IntegrationRunner(TestRunner):
+    """Runs the integration tests"""
+
+    def __init__(self):
+        """Initializes the failure tracker and test result map"""
+        TestRunner.__init__(self)
+        self.found_failure = False
+        self.result_map = {}
+
+
+    def add_options(self, parser):
+        """Add test-specific --version option"""
+        parser.add_argument("--version", dest="version", required=True,
+            help="The version that is expected to be installed, eg: '1.14.5'")
+
+    def run_test(self):
+        """Run all specified tests and inherit any failures"""
+
+        #List of tests to run
+        tests = [
+            [ "version", [ "--version", self.options.version ] ],
+        ]
+
+        for test in tests:
+            self.result_map[test[0]] = self.run_individual_test(test)
+
+        self.print_summary()
+
+        if self.found_failure:
+            sys.exit(1)
+
+    def run_individual_test(self, test):
+        """Run the actual test"""
+        command = [
+            "/usr/bin/env", "python3",
+            "-m", f"tests.integration.{ test[0] }",
+            "--platform", self.options.platform,
+            "--image", self.options.image,
+        ]
+
+        if len(test) > 1 and len(test[1]) > 0:
+            for arg in test[1]:
+                command.append(arg)
+
+        if self.options.verbose:
+            command.append("--verbose")
+
+        try:
+            output = subprocess.run(command, capture_output=True, check=True)
+        except subprocess.CalledProcessError as test_err:
+            self.found_failure = True
+            print_test_output(test[0], test_err.stdout, test_err.stderr)
+            return False
+
+        if self.options.verbose:
+            print_test_output(test[0], output.stdout)
+
+        return True
+
+    def print_summary(self):
+        """Print a summary to stdout"""
+        print("\n")
+        print(f"RESULTS: for { self.options.image } on { self.options.platform }")
+
+        successes = 0
+        failures = 0
+        for test, result in self.result_map.items():
+            if result:
+                successes += 1
+                result_str = "Success"
+            else:
+                failures += 1
+                result_str = "Failure"
+
+            print(f"{ test }: { result_str }")
+
+        sum_str = f"{ successes } successful tests and { failures } failures"
+        print(f"\nFinished test suite with { sum_str }")
+
+if __name__ == '__main__':
+    IntegrationRunner().main()
--- a/tools/genmatrix.js
+++ b/tools/genmatrix.js
@@ -6,7 +6,9 @@ const fs = require('fs');
 // we want to make decisions based on these changing
 const CI_FILES = [
   '.github/workflows/build-ci.yml',
-  'tools/genmatrix.js'
+  'tools/genmatrix.js',
+  'tests/integration',
+  'tests/integration_runner.py',
 ];
 
 // Regular expression for detecting version numbering of Dogecoin Core
@@ -128,6 +130,13 @@ class MatrixBuilder {
 
 }
 
+// Checks if any of the CI files have been changed
+const checkCIFilesChanged = (changedFiles) => {
+  return CI_FILES.some(file => changedFiles.some(changedFile => {
+    return changedFile.match(new RegExp(file));
+  }));
+}
+
 const generateMatrix = (baseDir, changedFiles) => {
 
   const builder = new MatrixBuilder(baseDir);
@@ -136,8 +145,9 @@ const generateMatrix = (baseDir, changedFiles) => {
   builder.readMatrix();
 
   // If the CI has changed, test all builds, otherwise, just build what's changed
-  const ciHasChanged = CI_FILES.some(file => changedFiles.indexOf(file) !== -1);
-  if (!ciHasChanged) {
+  if (checkCIFilesChanged(changedFiles)) {
+    console.log("CI files have changed, running all builds!");
+  } else {
     builder.filterIncludedVariants(changedFiles);
   }
 
--- a/tools/genmatrix.js
+++ b/tools/genmatrix.js
@@ -1,0 +1,153 @@
+'use strict';
+const path = require('path');
+const fs = require('fs');
+
+// Files that are used for CI scripts
+// we want to make decisions based on these changing
+const CI_FILES = [
+  '.github/workflows/build-ci.yml',
+  'tools/genmatrix.js'
+];
+
+// Regular expression for detecting version numbering of Dogecoin Core
+// used to detect which versions are what
+const VERSION_DIR_RE = /^\d+\.\d+\.\d+$/;
+
+// Entry in the matrix, resistant to red and blue pills
+class MatrixEntry {
+  constructor(version, variant, platform) {
+    this.version = version;
+    this.variant = variant;
+    this.platform = platform;
+  }
+
+  // returns the relative path to the version/variant as string
+  path() {
+    return path.join(this.version, this.variant);
+  }
+
+  // prints the entry to console
+  print() {
+    console.log(this.version, this.variant, this.platform);
+  }
+}
+
+// Build a GH Actions matrix
+class MatrixBuilder {
+
+  constructor(baseDir) {
+    this.baseDir = baseDir;
+    this.matrix = [];
+  }
+
+  // read a build matrix from the repository's directory hieararchy
+  readMatrix() {
+    // [1] read the versions from the repository root
+    const versions = this.readAllVersions();
+    if (!Array.isArray(versions) || versions.length < 1) {
+      throw new Error('No versions to iterate');
+    }
+
+    // iterate all versions
+    versions.forEach(version => {
+      // [2] read the variants from the version directory
+      const variants = this.readVariantsForVersion(version);
+      if (!Array.isArray(variants) || variants.length < 1) {
+        throw new Error('No variants to iterate for ' + version);
+      }
+
+      // iterate each variant for this version
+      variants.forEach(variant => {
+        // [3] Read the target platforms for this version/variant combination
+        const platforms = this.readPlatformsForVersionVariant(version, variant);
+        if (!Array.isArray(platforms) || platforms.length < 1) {
+          throw new Error('No platforms to iterate for ' + path.join(version, variant));
+        }
+
+        platforms.forEach(platform => {
+          // [4] Add an entry to the matrix for each version/variant/platform
+          //     combination
+          this.matrix.push(new MatrixEntry(version, variant, platform));
+        });
+      });
+    });
+  }
+
+  // read all subdirectories inside a given directory
+  // returns array of DirEnt objects
+  readDirs(fromDir) {
+    const files = fs.readdirSync(fromDir, { withFileTypes: true });
+    return files.filter(file => file.isDirectory());
+  }
+
+  // read all the version directories in the builder's baseDir
+  // returns array of filename strings
+  readAllVersions() {
+    const dirs = this.readDirs(this.baseDir);
+    const versionDirs = dirs.filter(dir => dir.name.match(VERSION_DIR_RE));
+    return versionDirs.map(dir => dir.name);
+  }
+
+  // read all variants for a specific version
+  // returns array of filename strings
+  readVariantsForVersion(version) {
+    const fromDir = path.resolve(this.baseDir, version);
+    const dirs = this.readDirs(fromDir);
+    return dirs.map(dir => dir.name);
+  }
+
+  // read all platforms for a specific version
+  // returns array of architecture strings, eg ["linux/amd64", "linux/arm64"]
+  readPlatformsForVersionVariant(version, variant) {
+    const pathToPlatformFile = path.resolve(this.baseDir, version, variant, 'PLATFORMS');
+
+    if (!fs.existsSync(pathToPlatformFile)) {
+      throw new Error('No PLATFORM file for ' + path.join(version, variant));
+    }
+
+    const contents = fs.readFileSync(pathToPlatformFile).toString('utf-8');
+    const platforms = contents.split('\n');
+
+    return platforms
+      .map(platform => platform.trim())     // remove whitespace just in case
+      .filter(platform => platform !== ''); // remove empty string entries
+  }
+
+  // filters the matrix against an array of files
+  filterIncludedVariants(fileList) {
+    this.matrix = this.matrix.filter(entry => {
+      const matchRegex = new RegExp(entry.path());
+      return fileList.some(file => file.match(matchRegex));
+    });
+  }
+
+  // output the expected GH Actions format
+  build() {
+    return this.matrix.length > 0 ? { include: this.matrix } : null;
+  }
+
+}
+
+const generateMatrix = (baseDir, changedFiles) => {
+
+  const builder = new MatrixBuilder(baseDir);
+
+  // populate the matrix with all builds
+  builder.readMatrix();
+
+  // If the CI has changed, test all builds, otherwise, just build what's changed
+  const ciHasChanged = CI_FILES.some(file => changedFiles.indexOf(file) !== -1);
+  if (!ciHasChanged) {
+    builder.filterIncludedVariants(changedFiles);
+  }
+
+  // print the build list for auditability.
+  console.log('BUILD LIST:');
+  builder.matrix.forEach(entry => entry.print());
+
+  // return the GH Actions expected matrix format as { "include": [Object] }
+  // or null if there are no entries
+  return builder.build();
+};
+
+module.exports = generateMatrix;


patricklodder:
I'm not entirely sure why this doesn't get triggered on this PR itself, so I'm investigating that. In the meantime, find a test run of this commit here: https://github.com/patricklodder/docker-dogecoin/actions/runs/1536909449

xanimo (commented):
works for me, can sus out the testing in another pr?
https://github.com/xanimo/docker/actions/runs/1537078369

patricklodder:
Fixed 3 issues:

1. Typo in js (`Regexp` -> `RegExp`)
2. Use `path.join()` instead of `path.resolve()` to omit a `./` when matching paths
3. Added `PLATFORMS` to CI triggers.

xanimo (dismissed):

patricklodder:
Linted with [semistandard](https://github.com/standard/semistandard) to standardize the js as much as possible but not run into readability issues because of all the piping going on (no semicolons will make it unreadable, so instead we just enforce semicolons).

1.14.5/bullseye/PLATFORMS
```diff
@@ -0,0 +1,4 @@
+linux/amd64
+linux/arm64
+linux/arm/v7
+linux/386
```
AbcSxyZ, line 4: Why do you need this file ? Can't hard code it into genmatrix.js ?

tools/genmatrix.js
```diff
@@ -0,0 +1,60 @@
...
+const CI_FILES = [
+  '.github/workflows/build-ci.yml',
+  'tools/genmatrix.js'
+];
+
+const VERSION_DIR_RE = /^\d+\.\d+\.\d+$/;
+
+const readVersions = (baseDir) =>
+  fs.readdirSync(baseDir, { withFileTypes: true })
+    .filter(f => f.isDirectory() && f.name.match(VERSION_DIR_RE))
```
AbcSxyZ: Can you use more meaningful variable than `f` please, `file` at least. For readability it's appreciable.

Same for the `o` variable in the log, I'm not even sure about what it is x)

AbcSxyZ (commented):
Quick review, I need to take time to dive into it and to see what's going on. I didn't do a lot of javascript and javascript nested calls within GitHub Action are fucking my brain...

Does the output of the `genmatrix.js` is a json with versions as keys and distros ? Variants are distros ? If you have an output example to share, it would be appreciated.

I will dive deeper in the PR in the following days, to be clear with the behavior.

1.14.5/bullseye/PLATFORMS
patricklodder: > Can't hard code it into genmatrix.js ?

from above:
> Either way, different Dogecoin Core versions and different distros can support different platforms, so we need to document this somewhere...

hardcoding is definitely not the answer.

AbcSxyZ: Fine, I missed it.

tools/genmatrix.js
patricklodder: Done!

1.14.5/bullseye/PLATFORMS
AbcSxyZ: Do you have an example ? If we have for example debian bullseye, bookworm and alpine, all support those 4 architecture. Could we have a different architecture where you want to release Dogecoin to ?
I'm wondering where we could have an issue

patricklodder:
> Does the output of the genmatrix.js is a json with versions as keys and distros ?

Yes, It's in GH Actions format for build matrices (`{ "include": [ KeyValuePair ] }`).

`genmatrix.js` is not an executable script, just a library, so that we can seamlessly integrate with GH Actions scripts to have sanitization rather than having to push everything through string transformations and hope for the best (otherwise, someone puts a space or unicode char in a filename and we'd be in trouble.) The GH Actions script calls `generateMatrix(baseDir, array-of-changed-file-paths-relative-to-the-repository-root)` from the library, which is currently the only exported function (the rest are internal functions.) Example output:

```js
{
  "include": [
    { 
      "version": "1.14.5",
      "variant": "bullseye",
      "platform": "linux/amd64",
    },
    ...
  ]
}
```

I kept `variant` from the node package instead of `distro` because we technically use it as a `distro`:`release`-`flavor` descriptor, so that seems to describe it the most inclusive.

patricklodder:
@AbcSxyZ after reading about headaches, I'll put this back to draft and instead of minimizing code, I'll write out 2 classes and make it more structured and documented so that it's easier to read. ETA ~ 3h.

patricklodder:
@AbcSxyZ let me know is this is easier to read. I haven't squashed the changes yet in case there's more.

AbcSxyZ:
I try to have an idea of the testing strategy to understand where we are going. I think it can depend on https://github.com/dogecoin/docker/issues/16 and a potential script to generate releases/updates.

I'm thinking about it because if we use templates, maybe the goal would be to be consistent across all versions, edit the template and propagate changes. I think it would be the ideal scenario, and in this case you will be sure every Dockerfiles/entrypoint will change simultaneously on updates.

How do you consider using a script to generate new releases & manage updates ?

> Different Dogecoin Core versions and different distros can support different platforms

Do you have any issue example ? Because if we go with `bullseye`, `bookworm` and in the future `alpine`, each support our 4 architectures used in the Dockerfile : `amd64`, `armv7`, `arm64` & `i386`.

If we stay with those distros with 4 type of binaries, I don't see the problem rn.

### Improvement suggestion

I was wondering how to test efficiently architectures. With this build test, if you give a fixed value for `ARCHITECTURE`, it won't fail. You can download the wrong binaries.

Maybe with the shasums file, we could grep hash by architecture (`RLS_ARCH`) instead of using the release file.

It's just an idea, not sure if it worth it and if we can't rely on the current Dockerfile for this, may ensure more accurately that with have downloaded the right file (at a potential cost of simplicity/readability).

patricklodder:
> I try to have an idea of the testing strategy to understand where we are going.

This is a Continuous Integration (CI) framework proposal. So what it enables is end-to-end integration tests, on PRs. #16 is a Continuous Deployment (CD) proposal, that runs on push. PRs precede push, and if you automate on push you better make sure that whatever you push is 100% correct, or else you'll push bugs. So for #16 to be feasible, we need to be sure that no PRs get merged that contain errors. This means that:

1. the build must succeed so that tests can be ran. That's what I implement here. It's just the framework to build the images that we have to test.
2. the resulting image must work. That's what we have to build a test suite on top of this PR, and check every image target against it - because each image is different.
3. the code must be clean. For this I will propose a linter, #29 is a step towards enabling that (so that we don't get errors on acceptance of the linter PR)
4. ideally, we know about any security issues so that we can fix whatever we can quickly, so we need a vuln scanner.

Note that none of this replaces any functional testing that you are proposing under #25.

> How do you consider using a script to generate new releases & manage updates ?

That imho only makes the need for a good CI greater, not smaller. The more you automate the better your checks must be. For now, feel free to write a bot that does PRs.

> Do you have any issue example ? Because if we go with bullseye, bookworm and in the future alpine, each support our 4 architectures used in the Dockerfile : amd64, armv7, arm64 & i386.

Incorrect, because you're only considering the current availability of only the docker bases. You need to look at 2 more variables at least: Dogecoin Core binary availability and time. Currently there are no Dogecoin Core binaries for musl-based linux. Looking at the process, it will be pretty hard to do musl compilation for different architectures all at once and I don't know how desirable that will be from a Dogecoin Core perspective. Say I get a working musl gitian descriptor for 1.21.0, then 1.14.5 still doesn't have it, but we will need to support that for a while. So over time, things will change. Will be the same for any other added architecture, or OS (think native docker for windows or darwin.)

> I was wondering how to test efficiently architectures.

Build the architecture, run the tests. Many images have a test stage, we can add this. That would have the upside that we can just build the test target(s) in buildx, but it will have a downside that these targets will get built when doing `docker build`, so I'm not 100% convinced of either. 

> With this build test, if you give a fixed value for ARCHITECTURE, it won't fail. You can download the wrong binaries.
>
> Maybe with the shasums file, we could grep hash by architecture (RLS_ARCH) instead of using the release file.

I have the feeling that you are proposing this because you don't like the level of automation I'm adding (or the `PLATFORMS` file) more than because there is something wrong with my proposal. But I didn't invent it, I stole the idea from a very high performing Official Docker image. 

As for the manifest I mentioned that I want instead of the `PLATFORMS` file, from a lifecycle management perspective I think this is a must-have. At the very least we must be able to disable automated builds due to having failures with an external source that we cannot fix, and we must be able to specify what is `:latest` and `:<major>`, all of that on a fine-grained level (image target) or else we will automate pushes with disastrous consequences. 

This is really just step 1 of many to get production quality images.

AbcSxyZ:
Multiples interrogations are coming to my mind due to multiple factors being involved, I'm definitively unsure about the strategy to adopt.

### Need of cross architecture build for testing ?

**At first, I'm not convinced it's useful to test cross architecture build, and I'm wondering if testing the build of single arch isn't enough to make sure it builds for all architectures ?**

I'm thinking about it for various reasons. We do not edit the Dockerfile related to architectures, it's a "built-in" features with `buildx`. Can't we assume if our Dockerfile build work for one architecture, it would work for every architecture ?

**Can you share the Docker repository where you get the design idea, please.** By looking at official images, they seem to perform CI on push for all versions, but do not test cross-architecture build (here for Github Action CI result for [php](https://github.com/docker-library/php/actions/runs/1540030931)). My previous idea could make sense regarding their testing mechanism.

Nor they mention information related to architectures in their [versions.json](https://github.com/docker-library/php/blob/master/versions.json). *They use Docker Hub to pull image and not GA, it  provides some testing tools too, maybe another layer of tests is done there but I doubt about it.*

Actually, we do not test that binaries are working on their architectures, and I suppose/assume it's done previously by Dogecoin Core tests. Maybe for the docker setup, we have more to take care about delivering the right executable per architecture rather than having executables running on those architectures. That was the idea behind my improvement suggestion, which can be an extra security check but does not in any way replace what we're implementing here.

Could it be enough to test the build for a single architecture and run tests from https://github.com/dogecoin/docker/pull/25 ?
I'm not against the automation, I'm thinking about the purpose :)

## Alpine strategy (and other OS)

I don't understand for Alpine what we could have. Will we have a binary released per architecture (with new arch added slowly), or would we build it from source (which is against gitian process as you may mention) ?

I'm wondering if we are not over-engineering for now. Definitively agree we should keep in mind extension for alpine or whatever when it's appropriate, but for now I feel like we want to anticipate a scenario where we would have different architectures between different version for some executables we don't even have.

Let's stay simple ? Write extensible code but do not add code for something who may come. Let's get a musl executable before implementing its Dockerfile.

## Update strategy based on template

If we choose to follow or not this path to work from a template to propagate changes, it could have important influence on the design from my pov. As I mentioned, some check would be avoided because if you update from a template, you're sure you would need to update all generated Dockerfiles. In some ways you do not work on edited files as you do here, but on all available versions tracked by an intermediary file (`versions.json` in official images).

Don't you think we should already consider moving to this template based strategy, to facilitate new releases, but also to facilitate maintaining (of various versions) ?

patricklodder:
> Can't we assume if our Dockerfile build work for one architecture, it would work for every architecture ?

No. I make a point about this because besides that Dogecoin Core itself does [not have quality gates](https://github.com/dogecoin/dogecoin/blob/31afd133119dd2e15862d46530cb99424cf564b0/.github/workflows/ci.yml#L65) defined for ARM-based releases other than "it compiles" right now (and this should be fixed there, but I'm still in the process of removing 1.14 technical debt - it's a seemingly endless work in progress), Debian [maintains a branch per architecture](https://github.com/debuerreotype/docker-debian-artifacts/branches/active) for the base image. That means we cannot (and should not) trust that any given debian image tag for architecture A yields the same results as architecture B. I am confident that the Debian image maintainers do their utmost best to make it awesome, but do not ever forget that this software governs a lot of value. The principle that crypto is built upon - per Bitcoin's lead - is after all: **don't trust, verify**. 

The question is of course how much will we verify. Making sure that each image works... is in my opinion not a big deal. It just means putting some work into automation. I've already done a lot of the work, click the "files changed" tab above. 

> Could it be enough to test the build for a single architecture and run tests from #25 ?

No, but it's a start. #25 doesn't implement any CI though. This framework could be used to automatically run the script from #25 cross-architecture.

> Maybe for the docker setup, we have more to take care about delivering the right executable per architecture rather than having executables running on those architectures.

The entire principle of continuous integration testing is built around learning about problems before you merge. Ignoring the fact that the architecture images are different can cause unforeseen issues. It's better to react to an issue when it's a PR, then when it's popping up in someone's production environment. The latter causes much stress and a costs a lot of time. The more inclusive the CI, the better we can prevent that.

> [..] they use Docker Hub to pull image and not GA [..]

CI should happen regardless of a pull or push publication. I think push can be more transparent, but I haven't studied Docker's publicly published push auditing in-depth. Maybe they have something better.

> Don't you think we should already consider moving to this template based strategy, to facilitate new releases, but also to facilitate maintaining (of various versions) ?

Yes. But it doesn't change anything. It basically removes the `PLATFORMS` file and puts it into a build manifest (what you call `versions.json`) and I already said in the first comment, and my last one, that we should absolutely go there. I just don't want to block CI with the work required to design a good manifest.  We need CI now, because we're only testing manually, and that's not good enough.

> Will we have a binary released per architecture (with new arch added slowly), or would we build it from source (which is against gitian process as you may mention) ?

For this repository, I think we **must** only release gitian-built binaries. Currently, one can create a user-managed Dockerfile for Alpine that self-compiles and the author takes full responsibility for any money lost while people are running it. It is imho not wise to do closed-process compilation from this organization. So yes, binaries must be released per gitian and yes, they will definitely not be for all architectures at once. I've seen the issues with the cross-compilation toolset for musl for x86_64 right now, and don't even want to think about ARM yet. One step at a time.

 > I'm wondering if we are not over-engineering for now.

Depends. We're basically adding deployment to something that never has done that before. With that comes a LOT of responsibility. I think we should take that seriously.

Question though: is all this just because you don't like the platform enumeration? The javascript? The fact that there is javascript? The fact that there is an extra file? All I hear is: let's not do it. Recent events in the Dogecoin ecosystem have lead to massive losses due to under engineering. Let's stay vigilant and not neglect verification that things work, because we're not a bank that can edit the books, or Ethereum for that matter.

AbcSxyZ:
>  is all this just because you don't like the platform enumeration? The javascript? The fact that there is javascript? The fact that there is an extra file? All I hear is: let's not do it.

Not at all. Sure, I would appreciate being able to use python to manipulate files with GitHub action :snake:, but they do not seem to offer this possibility. I'm not afraid of JavaScript, I just need friendly code :)

**I think my concern is the balance between utility of tests vs resources consumption**. It does not cost a lot of lines to launch a test for x4 architectures, but does it worth it and what are we testing ? Does the download/build of resources for versions * distros * architectures * pull request is not too big if we just verify that we can launch `apt` and `wget` on Debian for every platform ?

Actually, I wasn't even thinking about verifying Debian... I was thinking about verifying if executables worked, if executables are downloaded properly for different architectures, here it could make sense. As you mentioned, some tests are missing for arm, but it's not the place, and it should be done prior to this. We shouldn't run here the tests to ensure Dogecoin Core run on arm, so it's not what we're testing now.

Going in this logic (I'm not sure about what verifying Debian could mean), that's every executable where you put 0 trust, you wouldn't assume that `wget` can download a file on all platform, gpg to import keys, shasum to verify a checksum, bash to run an if condition and handle double quotes, and so on. How much download for this... which is really unlikely to happen. And if this day happen, when some of the Internet will break with us, maybe our push to Docker Hub will also break, avoiding the release of a broken image :muscle: 

For now, regarding our current tests and other official images repos (**could you link me the repo where you picked the idea ?**), I don't see the benefit of doing the build of cross architectures for testing I don't know what, and only see the important consumption of resources. So what are we testing which require cross compilation ?
That's my point, no problem with JavaScript  :dove:, no hate against any file :heart: !

patricklodder:
> I'm not afraid of JavaScript, I just need friendly code :)

I exploded 60 lines of lambdas into 150 lines of documented Builder pattern code... if you need more, tell me what. We already have a much more engineered version of what nodejs image does in a couple of lines... that I made for YOU.

> I think my concern is the balance between utility of tests vs resources consumption. 

How much resources does it consume then? Are you worried about the carbon footprint? Do I need to do it manually on each PR so that it costs renewable energy and my sweat only?

> Does the download/build of resources for versions * distros * architectures * pull request is not too big if we just verify that we can launch apt and wget on Debian for every platform ?

This really is not enough. Please point me to a CI that manages financial systems where checks go as far as checking if apt and wget works, that someone else packaged and tests... that gives zero meaning. The thing we want to test is if the image we propose to create with a new pull request works. And if it doesn't, we can fix it before merge.

 Currently, #25 only covers some functionality of entrypoint.py and doesn't test the Dockerfile other than that the entrypoint.py is copied there.  Just to give you an example based on the first line in that file: an error in `/usr/bin/env` or the renaming of `python3` into `python` on any architecture is not detected, because we're importing the file. This is fine, within the scope of that test.

> Actually, I wasn't even thinking about verifying Debian... I was thinking about verifying if executables worked

That's exactly what I'm proposing. Assuming that it works because apt and wget works, is not the same as verifying.

> We shouldn't run here the tests to ensure Dogecoin Core run on arm, so it's not what we're testing now.

Agreed. We should test that the binary that we've downloaded runs on the image we've created.

> you wouldn't assume that wget can download a file on all platform, gpg to import keys, shasum to verify a checksum, bash to run an if condition and handle double quotes, and so on

- If wget breaks, the build breaks.
- if gpg breaks, the build breaks.
- If shasum breaks, the build breaks.
- if bash is not handling quotes to compare architectures, the build breaks
- if the dogecoind binary we downloaded doesn't start up inside the container, **the build succeeds**. 

See my issue? The fact that wget, gpg, shasum and bash all work does not mean that it is safe to assume that the build is successful. That's what we need Continuous Integration for. And because each architecture has a different source (even if slightly) it is diligent to check them all if we can. And we can, because I coded that for us with this PR.

> could you link me the repo where you picked the idea 

- genmatrix principle comes from [docker-node](https://github.com/nodejs/docker-node/blob/b515bc8c80f6e20f789ee39a032888b26313579a/genMatrix.js)
- the idea to iterate through architectures comes from the [debian base image](https://github.com/debuerreotype/docker-debian-artifacts/blob/99a1a65c79b0d9d0cbc4dea18b9751deda9c991b/generate-stackbrew-library.sh#L9-L21)
- the `PLATFORMS` file is just a placeholder I created until we have a build manifest, as I explained [here](https://github.com/dogecoin/docker/pull/23#issue-1070923555), [here](https://github.com/dogecoin/docker/pull/23#issuecomment-986102107) and [here](https://github.com/dogecoin/docker/pull/23#issuecomment-987081703)

> So what are we testing which require cross compilation ?

The images, end-to-end. Every single one. We have 4 products, so we test all 4. If we have 20, we test all 20.

AbcSxyZ:
First, I didn't complain about current js code. If you did it only for me, go back to lambdas. If you did it to increase the readability even for non JavaScript dev in the goal of facilitating integration of people in the code, keep it. Don't do things specifically for ME because I'm too dummy, I just thought adding meaningful variable name and avoid too much nesting is appreciable.

>  Do I need to do it manually on each PR so that it costs renewable energy and my sweat only?

C'mon... There is a difference between asking someone to do test by hand vs doing right tests and avoid use of huge amount of resources for potentially useless tests. I'm not against doing it, I just need a meaningful purpose.

> If wget breaks, the build breaks.
> If gpg breaks, the build breaks.
> If shasum breaks, the build breaks.
> if bash is not handling quotes to compare architectures, the build breaks
> if the dogecoind binary we downloaded doesn't start up inside the container, the build succeeds.

So, how do we test the last step and not restrict ourselves to verify if wget & bash works ? For me, with the current PR, we test all steps without the last one. It's why I think it's a waste of resources.

I do not know any Dockerfile who cross build to just verify if the cross build works. Reliability of image like mysql is critical, more than us, they probably don't avoid the test of cross build only to get a more spicy life.

patricklodder:
>  If you did it to increase the readability even for non JavaScript dev in the goal of facilitating integration of people in the code, keep it.

Of course, I guess I put that too hard - if I wouldn't agree, I wouldn't have coded it. To me, it's a now versus later thing. I'm not saying you're dummy, at all. I'm trying to accommodate you in the proposal of a CI framework so that we can build upon it sooner rather than later, because currently we are manually testing and unless any of the reviewers has a consistent test framework that is ran consistently, we may be missing things. I'm not saying we have missed things, but this type of automation will help.

> avoid use of huge amount of resources for potentially useless tests

Like I mentioned in the last bullet point of my first comment, I left the tests out to focus on the framework first. So of course the framework is not complete. All it does now is test the build. It would have found the following PR comments:

- https://github.com/dogecoin/docker/pull/6#discussion_r758845462
- https://github.com/dogecoin/docker/pull/6#discussion_r761716899
- https://github.com/dogecoin/docker/pull/11#issuecomment-972159769

one out of 3 was architecture specific.

> So, how do we test the last step and not restrict ourselves to verify if wget & bash works ?

We need to test integration of the recipe for all resulting components.

I'd at the very least want to **run the container** with the following params:

- `-e VERSION=1` and no command
- `dogecoin-cli -version` as command
- `-e HELP=1` and `dogecoin-tx` as a command

that would allow us to verify that we got the correct binary, and make sure that the entrypoint works end-to-end.

Then, I think we must test `dogecoind` configuration (datadir!) and `dogecoin-cli` interaction by doing:

1. `dogecoind -regtest`
2. `dogecoin-cli -regtest generate 1`
3. `dogecoin-cli -regtest getblockinfo`
4. Ideally, we would verify network integration by both connecting to the p2p port and letting the dogecoind connect out.

And then we have a minimal integration test. I don't think this is complete, but it would be a good start.

> Reliability of image like mysql is critical, more than us

I fully disagree with this sentiment. MySQL is not a financial system so in case it would be used as a component in one, it simply passes on the test burden to the implementing system. I'm not sure that we can do that here.

From my personal experience in designing and building financial systems, CI for these systems are much more complex than what I'm proposing here and have many more checks than we can possibly produce. You seem to ignore that `dogecoind` is the single critical component in a decentralized timestamping infrastructure that guards billions, for millions of users. Billions that no one here controls or owns. And because its main function is timestamping, the decentralized ledger will move forward, even if one deployment is broken. For a database ledger, it's probably guarded by queues in the front, and you can patch up mistakes. That is not possible with this software. 

A block not interacted with cannot be reversed, and we have time-critical features, like [CLTV](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) and I'm quite sure we'll introduce [CSV](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki) as one of the next protocol enhancements. **Imagine being a merchant and your software is broken because we didn't catch a bug in our CI, and oops, a timelock expired while you were down... That can translate into real money lost.**

Small mistakes can be disastrous, especially when you realize that institutions are using this software for custodial services. Some, as recently became apparent, even use the wallet function to store billions of DOGE. This means that by publishing a software, this use-case is enabled. So we better make really sure that whatever we publish is good. Until the birth of this repository, deployment of dogecoind was not touched upon, so we're doing something pretty novel (and risky) here by taking this out of user space.

Yes, Dogecoin was made as a joke. But digressions aside, development of it has been a serious thing since development of 1.7. In any Dogecoin software published we have to be serious and precise, so that the joke can live on.

xanimo (dismissed):
ACK with rebase + squash of 'fixup'.

https://github.com/xanimo/docker/runs/4454292033?check_suite_focus=true

AbcSxyZ:
Mysql image is more critical than this image, nothing rely on this docker setup for now, where the opposite isn't true.  By being a fork, Dogecoin Core benefit of the work previously done by Bitcoin or Litecoin devs. Just taking this image and build critical infrastructure right now without further audit & work would be a mistake from this company, it's not comparable as using Dogecoin Core.

I hope what we are doing is improving the reliability, but this software need to be stressed & improved before being fully trustable. Dogecoin shows the power of jokes and fun, but yeah, it's not self-sufficient.

For me, it would be NACK without some sort of tests, at least `dogecoind -version`, to make the cross architecture build meaningful & to test something (right download, if binaries are running, whatever).

For the manifest/version file, I would say it's too early now, I would avoid it with this PR. Stay with 4 arch for Debian distros, and whenever we will have one or more musl executable (or more info about it), start to implement it at this moment if we need to manage dynamically architectures. It's mainly in the idea to stay simple.

The design will probably evolve, I guess there is important probability we won't use a `PLATFORM` file by folder for example, as you already mentioned.
It seems to me there is too much anticipation regarding alpine here.

## Test `dogecoind -version`

I tried to figure out some ways to do it. By adding an instruction `RUN dogecoind -version` during a cross architecture build, it can work properly:
```
#21 [final 10/10] RUN ls -l /usr/local/bin && /usr/local/bin/dogecoind -version
#21 0.055 total 54796
#21 0.055 -r-sr-xr-x 1 dogecoin dogecoin  3647544 Dec  8 14:07 dogecoin-cli
#21 0.055 -r-sr-xr-x 1 dogecoin dogecoin 37380240 Dec  8 14:07 dogecoin-qt
#21 0.055 -r-sr-xr-x 1 dogecoin dogecoin  4012448 Dec  8 14:07 dogecoin-tx
#21 0.055 -r-sr-xr-x 1 dogecoin dogecoin 11055032 Dec  8 14:07 dogecoind
#21 0.055 -r-x------ 1 root     root         4604 Dec  8 14:06 entrypoint.py
#21 0.059 Dogecoin Core Daemon version v1.14.5.0-31afd1331
#21 0.059 Copyright (C) 2013-2021 The Bitcoin Core and Dogecoin Core developers
#21 0.059 Please contribute if you find Dogecoin Core useful. Visit
#21 0.059 <https://dogecoin.com> for further information about the software.
...
```

Or crash on wrong architecture with the following error (amd executable on i386 image here): 
```
 > [linux/386 final 10/10] RUN ls -l /usr/local/bin && /usr/local/bin/dogecoind -version:
#26 0.219 total 54796
#26 0.219 -r-sr-xr-x 1 dogecoin dogecoin  3647544 Dec  7 22:49 dogecoin-cli
#26 0.219 -r-sr-xr-x 1 dogecoin dogecoin 37380240 Dec  7 22:49 dogecoin-qt
#26 0.219 -r-sr-xr-x 1 dogecoin dogecoin  4012448 Dec  7 22:49 dogecoin-tx
#26 0.219 -r-sr-xr-x 1 dogecoin dogecoin 11055032 Dec  7 22:49 dogecoind
#26 0.219 -r-x------ 1 root     root         4604 Dec  7 22:47 entrypoint.py
#26 0.220 /bin/sh: 1: /usr/local/bin/dogecoind: not found
------
Dockerfile:100
--------------------
  98 |     VOLUME ["/dogecoin/.dogecoin"]
  99 |     
 100 | >>> RUN ls -l /usr/local/bin && /usr/local/bin/dogecoind -version
 101 |     
 102 |     ENTRYPOINT ["entrypoint.py"]
--------------------
error: failed to solve: process "/bin/sh -c ls -l /usr/local/bin && /usr/local/bin/dogecoind -version" did not complete successfully: exit code: 127
Error: buildx failed with: error: failed to solve: process "/bin/sh -c ls -l /usr/local/bin && /usr/local/bin/dogecoind -version" did not complete successfully: exit code: 127
```

It somehow weird to get a not found error, I suppose it's coming from the fact it runs inside a container. I would expect an exec error. As follows when using amd image on arm :
```
pi@raspberrypi:~ $ sudo docker run abcsxyz/test:dogecoin-1
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm/v7) and no specific platform was requested
standard_init_linux.go:228: exec user process caused: exec format error
```
So we may assume a not found error = exec error 🤷

## Running tests from host or from image build/container
I'm not sure about the strategy to adopt, if tests need to be run from inside or outside the container/image.

To perform it from the host, GA do not offer natively [multi architecture hosts](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on), and suggest to use [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) which is not really possible. Idk if there is some viable alternative to get an arm host, I saw [Run-On-Arch GitHub Action](https://github.com/marketplace/actions/run-on-architecture) by quickly googling.

To perform it from the image, I made a test. What happen when doing `RUN dogecoind` in the Dockerfile. It works, and a node can be started during the build of an image, it's starts download and build steps are stuck here. I tried it locally for a single architecture, not for cross architecture on GA.

I discovered that build an image = running a container and launch Dockerfile commands. You can go inside the container used for the build for example :)
So we may perform some tests this way too. It may be possible to test some stuff using test mechanism from https://github.com/dogecoin/docker/pull/25, by calling executables.

Did you have some other ideas to implement this kind of tests ? What looks interesting to you ? I'm not sure if running tests inside docker + QEMU is reliable enough or if we can and/or should use multi architecture hosts.

patricklodder:
> By being a fork, Dogecoin Core benefit of the work previously done by Bitcoin or Litecoin devs.

Nice try. That's not a reason to be lax on end-to-end quality assurance. In fact, I think this is exactly the kind of thinking that causes bugs. If you're unsatisfied with the QA on Dogecoin, would you please join me in making it better rather than obstructing a CI framework PR?

> [..] nothing rely on this docker setup for now [..]
> Just taking this image and build critical infrastructure right now [..]

Keyword: **now**. I'm trying to get a framework in place so that the moment this hits publication (or even right now, this is public already) people will take this and do all the naughty things we would advise them to not do. And then lose money, possibly custodial money, when we screw up. I am not proposing this "for now", I'm proposing this to have a good foundation for the future. If you just want to play with Docker, then this is not the place.  

> For me, it would be NACK without some sort of tests

Alright. I'll code a test framework and a test on the top, inside this PR. Note that you're exploding scope on this PR and making review harder that way.

> Running tests from host or from image build/container

If you run it inside the container's build process, we're not using entrypoint like a user would, so that would not fully test integration of all the code in this repository. Let's try to cover the entire product (=image) in the integration test, and individual pieces in unit tests.

xanimo:
> ## Test `dogecoind -version`
> I tried to figure out some ways to do it. By adding an instruction `RUN dogecoind -version` during a cross architecture build...

Why don't we adopt what they do in ruimarinho/bitcoin-core and compare the output of `dogecoind -version` | grep the RLS_VERSION at the bottom of the file? 

```
 ​RUN​ bitcoind -version | grep ​"Bitcoin Core version v${BITCOIN_VERSION}"
```
which for us at least as of 1.14.3 appended daemon between core and version and use our RLS_VERSION like so:
```
RUN dogecoind -version | grep "Dogecoin Core Daemon version v${RLS_VERSION}"
```

This would be placed in between our ENTRYPOINT and CMD lines. Would this satisfy your testing requirement of `dogecoind -version`?

patricklodder:
- Test framework added ✅
  - it is written python ✅
  - it uses docker's capability to run cross-arch images ✅
  - it has the same structure as Dogecoin Core's integration tests ✅
  - it is extendable ✅
  - it tests every PR ✅
  - it tests every architecture ✅
  - it tests every changed version/variant for a PR ✅
  - it is the most awesome test framework you have ever seen written, yes it is ✅
- It is not complete 😱
  - because it's extendible, so it's easy to write more tests, feel free to write some ✅
- It is not *insert my favorite test framework* 😱
  - I have decided that retaining similarities with the Dogecoin Core repository is more important ✅
- This is a lot of code for just a version check 😱
  - that's because the framework is extendable and I will write more features in the future ✅
- Bitcoin and Litecoin don't do this 😱
  - I don't care ❤️

patricklodder:
> Why don't we adopt what they do in ruimarinho/bitcoin-core and compare the output of dogecoind -version | grep the RLS_VERSION at the bottom of the file?

Read the discussion above?

xanimo (dismissed):
ACK. Tested here https://github.com/xanimo/docker/actions/runs/1556480408 and locally:

linux/amd64
```
version
----------------------
Running command: docker run --platform linux/amd64 -e VERSION=1 xanimo/1.14.5-dogecoin:ci
Running command: docker run --platform linux/amd64 xanimo/1.14.5-dogecoin:ci dogecoin-cli -?
Running command: docker run --platform linux/amd64 xanimo/1.14.5-dogecoin:ci dogecoin-tx -?
Tests successful



RESULTS: for xanimo/1.14.5-dogecoin:ci on linux/amd64
version: Success

Finished test suite with 1 successful tests and 0 failures
```
linux/arm64
```
version
----------------------
Running command: docker run --platform linux/arm64 -e VERSION=1 xanimo/1.14.5-dogecoin:ci-arm64
Running command: docker run --platform linux/arm64 xanimo/1.14.5-dogecoin:ci-arm64 dogecoin-cli -?
Running command: docker run --platform linux/arm64 xanimo/1.14.5-dogecoin:ci-arm64 dogecoin-tx -?
Tests successful



RESULTS: for xanimo/1.14.5-dogecoin:ci-arm64 on linux/arm64
version: Success

Finished test suite with 1 successful tests and 0 failures
```
linux/arm
```
version
----------------------
Running command: docker run --platform linux/arm -e VERSION=1 xanimo/1.14.5-dogecoin:ci-arm
Running command: docker run --platform linux/arm xanimo/1.14.5-dogecoin:ci-arm dogecoin-cli -?
Running command: docker run --platform linux/arm xanimo/1.14.5-dogecoin:ci-arm dogecoin-tx -?
Tests successful



RESULTS: for xanimo/1.14.5-dogecoin:ci-arm on linux/arm
version: Success

Finished test suite with 1 successful tests and 0 failures
```
linux/386
```
version
----------------------
Running command: docker run --platform linux/386 -e VERSION=1 xanimo/1.14.5-dogecoin:ci-386
Running command: docker run --platform linux/386 xanimo/1.14.5-dogecoin:ci-386 dogecoin-cli -?
Running command: docker run --platform linux/386 xanimo/1.14.5-dogecoin:ci-386 dogecoin-tx -?
Tests successful



RESULTS: for xanimo/1.14.5-dogecoin:ci-386 on linux/386
version: Success

Finished test suite with 1 successful tests and 0 failures
```
lint:
```
pylint ./tests --verbose
No config file found, using default configuration

------------------------------------
Your code has been rated at 10.00/10
```

tests/integration_runner.py
```diff
@@ -0,0 +1,100 @@
...
+
+        for test in tests:
+            self.result_map[test[0]] = self.run_individual_test(test)
+
+        self.print_summary()
+
+        if self.found_failure:
+            sys.exit(1)
+
+        sys.exit(0)
```
AbcSxyZ: If you like:
```suggestion

        sys.exit(0 if not self.found_failure else 1)
```

tests/integration/framework/test_runner.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def main(self):
+        """main loop"""
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--platform", dest="platform", required=True,
+            help="The platform to use for testing, eg: 'linux/amd64'")
+        parser.add_argument("--image", dest="image", required=True,
+            help="The image or tag to execute tests against, eg: 'verywowimage'")
+        parser.add_argument("--verbose", dest="verbose", default=False, action="store_true",
+            help="The image or tag to execute tests against, eg: 'verywowimage'")
```
AbcSxyZ: Help menu not changed from the copy/paste.

tests/integration_runner.py
```diff
@@ -0,0 +1,100 @@
...
+    def add_options(self, parser):
+        """Add test-specific --version option"""
+        parser.add_argument("--version", dest="version", required=True,
+            help="The version that is expected to be installed, eg: '1.14.5'")
+
+    def run_test(self):
+        """Run all specified tests and inherit any failures"""
+
+        tests = [
+            [ "version", [ "--version", self.options.version ] ]
```
AbcSxyZ: You should end lists which are updated regularly with `,`, it facilitates versioning.
```suggestion
            [ "version", [ "--version", self.options.version ] ],
```

tests/integration/framework/docker_runner.py
```diff
@@ -0,0 +1,56 @@
...
+            print(f"Running command: { ' '.join(command) }")
+
+        try:
+            output = subprocess.run(command, capture_output=True, check=True)
+        except subprocess.CalledProcessError as docker_err:
+            print("STDERR:")
+            print(docker_err.stderr.decode("utf-8"))
+
+            print("STDOUT:")
+            print(docker_err.stdout.decode("utf-8"))
```
AbcSxyZ: Maybe outputs from stderr could be handled more appropriately across the tests. You're not using `sys.stderr` to display error message of the test, which are handled as stdout in the `IntegrationRunner`. It is somehow confusing.

It ends up with 3 forks IntegrationRunner -> VersionTest -> DockerRunner, I'm not sure if you really want to transform DockerRunner stderr to stdout by VersionTest.

A potential suggestion:
```suggestion
            print("$ " + " ".join(command), file=sys.stderr)
            print(docker_err.stdout.decode("utf-8"), file=sys.stderr)
            print(docker_err.stderr.decode("utf-8"), file=sys.stderr)
            sys.exit(1)
```
With an invalid image, I get:
```
versionx
----------------------
$ docker run --platform 386 -e VERSION=1 dogecoin-cross

Unable to find image 'dogecoin-cross:latest' locally
docker: Error response from daemon: pull access denied for dogecoin-cross, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.





RESULTS: for dogecoin-cross on 386
version: Failure
versionx: Failure

Finished test suite with 0 successful tests and 2 failures
```
With the idea of stdout print before stderr to reproduce the docker command (if some previous stderr messages are not messing the output...).

Possibly some improvement related to outputs if you didn't avoid stderr intentionally.

tests/integration_runner.py
```diff
@@ -0,0 +1,100 @@
...
+import subprocess
+import sys
+
+from .integration.framework.test_runner import TestRunner
+
+class IntegrationRunner(TestRunner):
+    """Runs the integration tests"""
+
+    def __init__(self):
+        """Constructor"""
```
AbcSxyZ: Pylint required a comment ?
```suggestion
```

tests/integration/version.py
```diff
@@ -0,0 +1,47 @@
...
+        self.ensure_version_on_first_line(dogecoincli.stdout)
+
+        # check dogecoin-tx
+        dogecointx = self.run_command([], ["dogecoin-tx", "-?"])
+        self.ensure_version_on_first_line(dogecointx.stdout)
+
+    def ensure_version_on_first_line(self, cmd_output):
+        """Assert that the version is contained in the first line of output string"""
+        first_line = cmd_output.decode("utf-8").split("\n")[0]
+        assert re.match(self.version_expr, first_line) is not None
```
AbcSxyZ: I failed a test by replacing `f".*{ self.options.version }.*"` by `f".*{ self.options.version }69.*"`, outputs are really raw:
```
version
----------------------
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/abcsxyz/Documents/dogecoin/docker/tests/integration/version.py", line 47, in <module>
    VersionTest().main()
  File "/home/abcsxyz/Documents/dogecoin/docker/tests/integration/framework/test_runner.py", line 51, in main
    self.run_test()
  File "/home/abcsxyz/Documents/dogecoin/docker/tests/integration/version.py", line 31, in run_test
    self.ensure_version_on_first_line(dogecoind.stdout)
  File "/home/abcsxyz/Documents/dogecoin/docker/tests/integration/version.py", line 44, in ensure_version_on_first_line
    assert re.match(self.version_expr, first_line) is not None
AssertionError

Running command: docker run --platform amd64 -e VERSION=1 dogecoin-cross
```
Test crash, but debug may not be really friendly, no idea of outputs, required versions, tested values. I suppose it would be the same for all AssertionError, could be potential improvement.

I didn't see the output myself, but I saw that BitcoinTestFramework is doing some tricks related to AssertionError :
https://github.com/dogecoin/dogecoin/blob/31afd133119dd2e15862d46530cb99424cf564b0/qa/rpc-tests/test_framework/test_framework.py#L150

But, it seems to just reprinting the exception. Pytest are able to display value, but with a quick view of the code, they seem to rewrite assert instructions and are playing with the python ast... Some magic tricks, I ran away !

Work to handle errors, would be great to improve assert handling for a test framework for the debug, but not sure how rn.

AbcSxyZ (commented):
We maybe can do it in 2 steps, move the test framework inside a new PR and add the GA integration right after.

I comment here for now, I dive into the code it's a first review. It works fine, as I see, we will use qemu emulation finally.

The framework require some engineering to run (vs a framework where it's already done), it may require some time to appropriate it [doc needed], test failures do not have real debug information.

So, is it the *"most awesome test framework you have ever seen written"* ? Sadly ... I may break a dream ... but I don't think so :speak_no_evil:.
I would love to see it, but it's not for today :heart: 

Tests could be build on top of existing framework (this kind of test could have been implemented on top of #25, inside a container, using pytest for example), but I clearly agree of the importance to be consistent with Dogecoin Core test framework. Could we use a simpler test framework because needs with the Core are different, or is it better to stay with a design like the Core ? It's a matter of choice and it's up to you, I will follow the decision.

If we go this way, we clearly have to break #25 and change the design of the tests, otherwise we will have 2 concurrent testing mechanism.

tests/integration/version.py
patricklodder: I think that inside the version test I can write a better assertion than the bare one that I currently do, to make that display better. Will propose an improvement.

tests/integration_runner.py
patricklodder: Actually, I forgot to replace my placeholder, will amend.

tests/integration/framework/docker_runner.py
patricklodder: I'll rework this.

tests/integration_runner.py
patricklodder: That does in my opinion not increase readability. I've been trying to make everything as readable as possible per your earlier objections to readability of lambdas, so I'd prefer to keep this clean. I may have a refactoring idea on this though, please bear with.

patricklodder:
> We maybe can do it in 2 steps, move the test framework inside a new PR and add the GA integration right after.

Yesterday you said "For me, it would be NACK without some sort of tests, at least `dogecoind -version`". I have facilitated this request, so this is it for now. If you have found something new to NACK on, please say so and explicitly mention the blocker, so that I can consider fixing it. 

> [..] test failures do not have real debug information.

Addressing this, but can we please make this subject to future improvement?

> Sadly ... I may break a dream ... but I don't think so 🙊.

Way ahead of you! I already anticipated and answered your opinion per the second part of my original sentence 😂 - it was already self-deprecation, no need to deprecate me, I've already done it for you. Thank you for granting me the opportunity to improve though!

> Could we use a simpler test framework because needs with the Core are different, or is it better to stay with a design like the Core ?

This will move closer to features offered by Core's framework when we write tests that interact with a running container. DockerRunner will then become stateful for starting and stopping containers. I could have simply made it a function call now and skip class instantiation, but this is anticipating the greater scheme of where we need to go. I removed a lot of methods from my draft class, to ease review now, but will introduce it when it's needed. Make your life easy by doing a step-by-step approach while planning for the future.

> If we go this way, we clearly have to break #25 and change the design of the tests, otherwise we will have 2 concurrent testing mechanism.

We will have more than 2 concurrent testing/auditing mechanisms; right now proposed:

1. **Unit tests** of code maintained here (currently only entrypoint.py and the Dockerfile) as you propose in #25
2. A test framework for automated **integration testing** (this PR)

Both are needed! I beg you to please not refactor #25 to match this because it has a completely different purpose: unit tests validate atomic function behavior against its technical design and need features like mocking and code-traceability, whereas integration tests validate the behavior of the entire product assembly against its functional design, and need execution log traceability. Integration test failures are inherently more difficult to troubleshoot, and I support your remarks that the trace is not as good as it can be, yet.

But then, as I mentioned earlier, we will also need:

3. We need linters, at least pylint for entrypoint.py, and something like [hadolint](https://github.com/hadolint/hadolint) for the Dockerfile, to enforce standards. (We may also want to at least document debt by linting the test and build code, maybe in a non-blocking way - I have not entirely formed a full opinion yet, but something should be done)
4. We need something to do security audits on built images, I haven't taken the time to find a good o/s auditor yet.

and then finally:

5. Continuous deployment with automatic build, merge error prevention and & publish per #15
6. We probably need dependabot to keep GH actions up-to-date and free of known bugs/vulnerabilities in the CI/CD pipeline itself.

tests/integration/version.py
patricklodder:
```
$ python3 -m tests.integration_runner --platform linux/amd64 --image 4b3c1078692b --version 1.14.4


version
----------------------
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/patrick/docker-dogecoin/tests/integration/version.py", line 60, in <module>
    VersionTest().main()
  File "/home/patrick/docker-dogecoin/tests/integration/framework/test_runner.py", line 51, in main
    self.run_test()
  File "/home/patrick/docker-dogecoin/tests/integration/version.py", line 31, in run_test
    self.ensure_version_on_first_line(dogecoind.stdout)
  File "/home/patrick/docker-dogecoin/tests/integration/version.py", line 57, in ensure_version_on_first_line
    raise AssertionError(text)
AssertionError: Could not find version 1.14.4 in Dogecoin Core Daemon version v1.14.5.0-31afd1331




RESULTS: for 4b3c1078692b on linux/amd64
version: Failure

Finished test suite with 0 successful tests and 1 failures
```

AbcSxyZ:
Fine to avoid a separate PR, it was potentially to speak about the handling of error message/assert errors + implication of the framework with #25.

Tests are working, nothing is really blocking for the test framework.

About the dynamic management of architecture through `PLATFORM` file, do you think it's really needed ? We probably will move to a centralized manifest, it's designed for alpine which is not here. Important probability to not be used, stay with 4 Debian archs ?

Idk when we would introduce manifest, maybe for the templating it could be useful, for qa at a moment, but I think this PR may be not the right place nor the right design.

Actually, the introduction of a manifest will simplify genmatrix.js behavior.

patricklodder:
> About the dynamic management of architecture through PLATFORM file, do you think it's really needed ? We probably will move to a centralized manifest, it's designed for alpine which is not here. Important probability to not be used, stay with 4 Debian archs ?

Alternatively we can hardcode it inside the genmatrix.js file in the interim, which I think is worse than this proposal, because changing that file indicates a CI change and will remove benefits of using resources sparingly; see [this test pr CI run](https://github.com/patricklodder/docker-dogecoin/runs/4475496229) as an example of how the matrix generation optimizes the test as described in bullet point 3 of my original comment all the way up there...

I'm okay with replacing it but I'd suggest doing it later, especially since we have more pressing issues right now, like getting the CI and linters up, so that PRs see QA and we have some assurances that we don't create semi-functional images and have to do a lot of error-prone manual testing when we think this is ready for a first release.

> Idk when we would introduce manifest

Doesn't have to be now, but we either do it before we release this and have a future-proof build in place, or do it before we are faced with a new version of either Debian or Dogecoin Core. I can tell you that unless larger CVEs get reported, current work on 1.14.6 will take some time due to the security enhancement requested yesterday.

> Actually, the introduction of a manifest will simplify genmatrix.js behavior.

I'm not able to guarantee simplification completely (do strife for it) but I agree that it will have to change, for sure. We probably need to add things to it for #15 too.

tests/integration_runner.py
patricklodder: I found that the sys.exit(0) was redundant, so I removed it.

tests/integration/framework/docker_runner.py
patricklodder: reworked to consistently keep stderr in stderr, and stdout in stdout.

patricklodder:
Please let me know if the assertion improvement in version.py is good enough for now. If it is, I'll squash the fixups

.github/workflows/build-ci.yml
```diff
@@ -0,0 +1,80 @@
...
+on:
+  pull_request:
+    paths:
+      - "**/Dockerfile"
+      - "**/docker-entrypoint.py"
+      - "**/PLATFORMS"
+      - "tests/"
+      - "tools/genmatrix.js"
+      - ".github/workflows/build-ci.yml"
+
```
AbcSxyZ, line 12: For main only ? While we don't have a branch strategy
[warning indentation, to copy manually if used]
```suggestion
      - ".github/workflows/build-ci.yml"
    branches:
      - main         
```

AbcSxyZ (dismissed):
ACK for me, test working locally and with [GA](https://github.com/AbcSxyZ/docker/pull/6). For the assert, it's sad to not use `assert` keyword, but I don't see how it could be better for now.

Similarly, you could have used the following, not sure if the result is exactly the same or slightly different:
```python
text = f"Could not find version { self.options.version } in { first_line }"
assert re.match(self.version_expr, first_line) is None, text
```
But for clarity, I prefer your version.

This sentence is unclear for me *"that file indicates a CI change and will remove benefits of using resources sparingly;"*. I still don't see the real utility of `PLATFORM` and in favor of hard coded values, but if you prefer to remove it later.

.github/workflows/build-ci.yml
```diff
@@ -0,0 +1,80 @@
+name: build-ci
+
+on:
+  pull_request:
+    paths:
+      - "**/Dockerfile"
+      - "**/docker-entrypoint.py"
```
AbcSxyZ:
```suggestion
      - "**/entrypoint.py"
```

.github/workflows/build-ci.yml
patricklodder: This would limit the CI trigger to exclusively apply for pull requests that are raised with `main` as the base. Even if we would have more branches than `main` in this repository - I'd rather not - would we not want this to trigger when a PR is filed against, say, a split-off development branch? If not, why not?

AbcSxyZ: Yes, not necessary a good idea

AbcSxyZ (dismissed):

patricklodder:
I'll squash now for final review on a clean commit set, because this has too many useless commits right now

AbcSxyZ (approved):
ACK, tested locally on amd64 & with [GA](https://github.com/AbcSxyZ/docker/actions/runs/1560776373). Nice job, a good step :)

xanimo (approved):
ACK. Tested locally and on [GA](https://github.com/xanimo/docker/actions/runs/1560995553). Good catch on the `docker-entrypoint.py` -> `entrypoint.py` in .github/workflows/build-ci.yml @AbcSxyZ! And great job @patricklodder! Thank you both!

PR #26, opened 2021-12-04
Create main function in entrypoint + rename entrypoint

AbcSxyZ:
Create a main function to call in entrypoint + rename `docker-entrypoint.py`

Facilitate import of entrypoint for test coverage in https://github.com/dogecoin/docker/pull/25.

PR #27, opened 2021-12-04
Remove arguments from environment variables

AbcSxyZ:
Reimplement https://github.com/dogecoin/docker/pull/10 which have been removed by https://github.com/dogecoin/docker/pull/11, merged without rebase.

### DIFF
--- a/1.14.5/bullseye/docker-entrypoint.py
+++ b/1.14.5/bullseye/docker-entrypoint.py
@@ -99,7 +99,7 @@ def convert_env(executable):
 
     cli_arguments = []
     for option in man_options:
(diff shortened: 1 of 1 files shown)


patricklodder:
> merged without rebase.

Github should have given a merge issue then (but I didn't force merge.)

I will enforce rebase as a rule as soon as the CI from #23 is in.

patricklodder (approved):
ACK - code reviewed and tested on x86_64

patricklodder:
I'm merging this one now since we already ACK'd on #10, and explicitly before #26 because that needs to be reworked.

Issue #22, opened 2021-12-03
Optimize the build order


patricklodder:
_Originally posted by @AbcSxyZ in https://github.com/dogecoin/docker/pull/6#discussion_r759717146_

Separation of declaration and execution is not the rule in Dockerfile.

For exemple, [php](https://github.com/docker-library/php/blob/master/8.0/buster/apache/Dockerfile) & [golang](https://github.com/docker-library/golang/blob/master/1.17/bullseye/Dockerfile) are mixing declaration and execution. And probably all other Docker official images.

It's even a recommendation to use the cache with more efficiency, from [cacheability](https://github.com/docker-library/official-images#cacheability) recommendation:
> Ensure that lines which are less likely to change come before lines that are more likely to change (with the caveat that each line should generate an image that still runs successfully without assumptions of later lines).
>
> For example, the line that contains the software version number (`ENV MYSOFTWARE_VERSION 4.2`) should come after a line that sets up the APT repository `.list` file (`RUN echo 'deb http://example.com/mysoftware/debian some-suite main' > /etc/apt/sources.list.d/mysoftware.list`).

Doing all declarations at the top of a Dockerfile is defeating the utility of caching mechanism.

patricklodder:
I'll do an analysis after #20

patricklodder:
Below my initial analysis, I kept your suggestion of doing architecture detection together with verification. That will not improve readability of the latter, but can be done like that as long as we do not need architecture for any other decisions in the process. 

### Verify stage

| Step | Description     | Depends on       | Reposition | Rationale
| ---: | :-------------- | :--------------- | ---------: | :--------
|    1 | Workdir         | (static)         |          1 |
|    2 | RLS_VERSION     | (static)         |          5 | caught in dir structure
|    3 | RLS_OS          | (static)         |          6 |
|    4 | RLS_LIB         | 3                |          7 | tied to OS
|    5 | RLS_ARCH        | -                |          8 | 
|    6 | Architecture    | 4,5              |          - | combine with 16
|    7 | SIG_PATH        | 2,3              |         10 |
|    8 | DESCRIPTOR_PATH | 3                |         11 |
|    9 | RLS_LOCATION    | 2                |         12 |
|   10 | REPO_GITIAN_BUI | (static)         |          2 | can be on top
|   11 | REPO_GITIAN_SIG | (static)         |          3 | can be on top
|   12 | REPO_DOGECOIN_C | (static)         |          4 | can be on top
[...]

PR #31, opened 2021-12-08
Optimize the build order to use cache better.

patricklodder:
Partially implements #22

Verify:
- Put static(ish) repository locations at the top because those do not change often
- Move shasum pins to be right after version/variant declaration because they change when those change
- Merge architecture detection into the verification step to save an execution step and remove the build environment portability hack

Final:
- install python as late as possible, before copying entrypoint to stay more developer friendly for now. May need to be changed from a security perspective.
- move the entrypoint inclusion as far down the process as we can

### DIFF
--- a/1.14.5/bullseye/Dockerfile
+++ b/1.14.5/bullseye/Dockerfile
@@ -2,35 +2,27 @@ FROM debian:bullseye-slim AS verify
 
 WORKDIR /verify
 
+# github repository locations
+ARG REPO_GITIAN_BUILDER=https://github.com/devrandom/gitian-builder.git
+ARG REPO_GITIAN_SIGS=https://github.com/dogecoin/gitian.sigs.git
+ARG REPO_DOGECOIN_CORE=https://github.com/dogecoin/dogecoin.git
+
 # Specify release variables
 ARG RLS_VERSION=1.14.5
 ARG RLS_OS=linux
 ARG RLS_LIB=gnu
 ARG RLS_ARCH=
 
-# Automatically detect architecture
-RUN set -ex && ARCHITECTURE=$(dpkg --print-architecture) \
-    && if [ "${ARCHITECTURE}" = "amd64" ]; then RLS_ARCH=x86_64 ; fi \
-    && if [ "${ARCHITECTURE}" = "arm64" ]; then RLS_ARCH=aarch64; fi \
-    && if [ "${ARCHITECTURE}" = "armhf" ]; then RLS_ARCH=arm && RLS_LIB=gnueabihf; fi \
-    && if [ "${ARCHITECTURE}" = "i386" ]; then RLS_ARCH=i686-pc; fi \
-    && if [ "${RLS_ARCH}" = "" ]; then echo "Could not determine architecture" >&2; exit 1; fi \
-    && RLS_FILE_NAME=dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz \
-    && echo -n ${RLS_FILE_NAME} > .filename
-
-ARG SIG_PATH=${RLS_VERSION}-${RLS_OS}
-ARG DESCRIPTOR_PATH=dogecoin/contrib/gitian-descriptors/gitian-${RLS_OS}.yml
-
-ARG RLS_LOCATION=https://github.com/dogecoin/dogecoin/releases/download/v${RLS_VERSION}
-ARG REPO_GITIAN_BUILDER=https://github.com/devrandom/gitian-builder.git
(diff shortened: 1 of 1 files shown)


patricklodder:
Draft until we can test this.

patricklodder:
This is ready now. Good test that the CI process works. I'm going to get some PRs in now.

xanimo (dismissed):
ACK.

1.14.5/bullseye/Dockerfile
```diff
@@ -97,5 +88,13 @@ EXPOSE 22555 44555 18332
 
 VOLUME ["/dogecoin/.dogecoin"]
 
+COPY entrypoint.py /usr/local/bin/entrypoint.py
+RUN chmod 500 /usr/local/bin/entrypoint.py
+
+# Dependencies install
+RUN apt update && apt install -y \
+    python3 \
+    && rm -rf /var/lib/apt/lists/*
```
AbcSxyZ: Annoying for development, will reinstall python at each modification of entrypoint. Why would you do that ?

patricklodder: Documented in my comment earlier in https://github.com/dogecoin/docker/issues/22#issuecomment-988991126:

> *After thinking about this more, I think it's better when an entrypoint change invalidates the cache of the apt install of python, as a cached version may have a python that does not work with the entrypoint code the same that the latest python from the package maintainers repository. This means that if entrypoint changes, it MUST be built (and tested) with the latest python, because otherwise cached build steps can yield different results than non-cached builds ran at the same point in time.*

So, this is a classic security vs efficiency problem. We can turn it back but then we will have to document explicitly that images should be built with `--no-cache` for production deployments (and for any published images.) 

I am going to have to do some research to see how exactly that impacts deployment because I fear that this would eradicate incremental downloads. So where we would save `<number of local builds per release>` of pressure towards apt repositories, we would add `<number of deployments>` of pressure to Docker Hub. The latter is greater than the former at some point or we'd be developing this mostly for the sake of developing, not usage.

I think if we want the best of both worlds, we have to stop using apt and pin a specific python version, then maintain that. At that point, we control the caching. Cost: sweat and tears. I don't think this repository currently has enough participants to really offer continuity for that, but we can maybe plan for something.

*Note: From my personal experience, docker's caching is great for rapid development, but complicates security of deployment. I have on several occasions done development on a separate `Dockerfile.develop` that just optimizes cache by taking changed files last, and use a distinct Dockerfile for production that optimizes for cache invalidation and security. Also, I have significantly reduced my personal reliance on docker to test code under development in favor of local tools in my development environments, to save surprises later.*

patricklodder: Reverted for now. Will do security review for deployment later.

AbcSxyZ (commented):
As I understand Docker recommendations, the ideal may be:
```
# Apt install
# Release variables / repository location / static derivated variables
# Clone gitian builder/keys/dogecoin repo
# Shasum variables generation
# Download executables
```
They usually use apt before variables related to the application, which may change more easily when you develop/edit the content. I understand it like the variable declaration should come as late as possible for a `RUN` instruction. It's how I understand the standard.

A lot of different situations, and "optimal cache" may really vary of the use, regarding which variable can change. Those Dockerfiles are really designed for a specific version, a lot of things may not change, difficult to know...

In some ways it looks really nice, but with variable defined perhaps too much at the top.

However you want, I'm fine with everything :) We can add suggested changes to follow some recommendations (as I understand it), or keep a more elegant solution with (potentially) fixed variables at the top.

patricklodder:
So, in the 7.5 years release history of Dogecoin there has never been a gitian rebuild after release publication. So shasums, versions, and so on, are much more statically linked to the version/variant/Dockerfile than with a Dockerfile that compiles from source, and definitely changes less than the debian-packaged python, which has seen 5 CVEs this year so that would have been at least 5 changes to, say, a 1.14.3 image if we'd done it back then.

This problem will resolve itself shortly because hadolint will force us to pin versions and do the blood, sweat & tears work of maintaining the dependencies precisely - unless we suppress it, but I'd rather go with the pinning and do the work like a smarty shibe. We just need to find a good way to alert the repo of a needed change, maybe dependabot if it can help us with this. It's on my list to look into as soon as we get through the current series of proposed Dockerfile changes, including this PR.

That said, we also need to pin gitian builder somehow 😕

AbcSxyZ (approved):
I agree, it's really fixed variables we may rarely or never change. But it's because it's variables x)

Go for it, it's pretty nice (for me) too.
ACK.

xanimo (approved):
this looks good to me and ofc we'll always have time for improving/optimizing things in the future. for now though this looks and works good for me. ACK.

PR #32, opened 2021-12-10
qa: pick a js standard and just comply

patricklodder:
Easier to just comply to a standard and make it all awesome.

Standard: semistandard - because without semicolons function piping becomes hard to decipher. 

Changes made to comply with the standard:
  - remove trailing comma
  - space between class method name and input signature
  - no block padding
  - no multi-space to align comments
  - semicolons everywhere

### DIFF
--- a/tools/genmatrix.js
+++ b/tools/genmatrix.js
@@ -8,7 +8,7 @@ const CI_FILES = [
   '.github/workflows/build-ci.yml',
   'tools/genmatrix.js',
   'tests/integration',
-  'tests/integration_runner.py',
+  'tests/integration_runner.py'
 ];
 
 // Regular expression for detecting version numbering of Dogecoin Core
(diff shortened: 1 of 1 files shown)


AbcSxyZ (approved):
ACK. I was wondering if GA was using pushed code, it was interesting to notice it is using workflow & genmatrix.js of the PR.

You will have a linter to add later ?

patricklodder:
> You will have a linter to add later ?

I have linters ready for both py and js, am still figuring out hadolint

xanimo (approved):
ACK. however with regards to semi-colons everywhere, is this not applicable to classes?

patricklodder:
>  is this not applicable to classes?

not `class {}` and not `function {}` because those are declarations, also not things like blocks `if (test) { }`. But `const a = function(){};` is an expression so there yes.

PR #34, opened 2021-12-10
use executables -help menu instead of man page to get options

AbcSxyZ:
Fix #28.  Needed also for test coverage of Dockerfile in #25.

Use of `-help` menu to get option instead of man pages.

**Do you think the regex from grep is safe : `^  -'` ?** It handles all options with 2 space from the start of the line, do you know if it's done like that ?  I started with `^ *-`, but some options may use others in their description like `-listen` or `-checkblockindex`.

What would you like to hook from here ? Do we stay with grep, or we could break the grep and use regex from python to generate a more raw entry easier to hook ? The grep option may be nicer but harder to test.

With the grep, we may hook with something like this:
```python
dogecoind_menu = [
    "  -blockmintxfee=<amt>",
    "  -datadir=<dir>",
    "  -server",
    ...
]
dogecoin_cli_menu = [
    "  -datadir=<dir>",
    ...
]
```
Or if we want to test also the regex on a raw file, we break it and return some raw string like the help menu.

Because it uses executables, a test from #25 is breaking for `dogecoin-qt` with the following:
```
E               subprocess.CalledProcessError: Command 'dogecoin-qt -help -help-debug | grep '^  -'' returned non-zero exit status 1.

/usr/lib/python3.9/subprocess.py:528: CalledProcessError
------------------------------------- Captured stderr call --------------------------------------
dogecoin-qt: error while loading shared libraries: libxkbcommon-x11.so.0: cannot open shared object file: No such file or directory
```
Issue related to X window system, would be great to have a discussion related to `dogecoin-qt` as you mentioned previously, to see how/if we handle it.

PR #35, opened 2021-12-11
refactor: removed dogecoin-qt

xanimo:
removed dogecoin-qt from Dockerfile and entrypoint.py since it's not supported and wasting resources.

refined Dockerfile by extracting the specific files we need from our archive and copying each while setting appropriate permissions and owner:group in each COPY.

removed unnecessary WORKDIR /tmp since we are only copying specific files from verify build stage, thus we have nothing to rm -rf *

### DIFF
--- a/1.14.5/bullseye/Dockerfile
+++ b/1.14.5/bullseye/Dockerfile
@@ -72,8 +72,7 @@ COPY --from=verify /verify/dogecoin.tar.gz ./
 # Move downloaded binaries and man pages in the container system.
 # Setuid on binaries with $USER rights, to limit root usage.
 RUN tar -xvf dogecoin.tar.gz --strip-components=1 \
-    && cp share/man/man1/*.1 /usr/share/man/man1 \
-    && cp bin/dogecoin* /usr/local/bin \
+    && cp bin/dogecoind bin/dogecoin-cli bin/dogecoin-tx /usr/local/bin/ \
     && chown ${USER}:${USER} /usr/local/bin/dogecoin* \
     && chmod 4555 /usr/local/bin/dogecoin* \
     && rm -rf *
--- a/1.14.5/bullseye/entrypoint.py
+++ b/1.14.5/bullseye/entrypoint.py
@@ -11,7 +11,6 @@ import subprocess
 
 CLI_EXECUTABLES = [
         "dogecoind",
-        "dogecoin-qt",
         "dogecoin-cli",
         "dogecoin-tx",
         ]
@@ -53,7 +52,7 @@ def executable_options(executable):
     command_arguments = [executable, "-help"]
 
     #`-help-debug` display extra flag in help menu for dogecoind & qt
-    if executable in ["dogecoind", "dogecoin-qt"]:
+    if executable == "dogecoind":
         command_arguments.append("-help-debug")
 
     help_options = get_help(command_arguments)
@@ -127,7 +126,7 @@ def run_executable(executable, executable_args):
     to manage a single process in a container & more predictive
     signal handling.
     """
-    if executable in ["dogecoind", "dogecoin-qt"]:
+    if executable == "dogecoind":
         executable_args.append("-printtoconsole")
 
     #Switch process from root to user.


xanimo:
> removed dogecoin-qt from Dockerfile and entrypoint.py since it's not supported and wasting resources.
> 
> refined Dockerfile by extracting the specific files we need from our archive and copying each while setting appropriate permissions and owner:group in each COPY.
> 
> removed unnecessary WORKDIR /tmp since we are only copying specific files from verify build stage, thus we have nothing to rm -rf *

in a sub optimal fashion of not submitting PR's with atomic commits, i forgot to mention changing python3 to python3-minimal and adding .gitignore and .dockerignore files as placeholders to remind us that they're excellent assets to reduce overhead on system resources in particular build context archives every time we run build.

this reduces the linux/amd64 image for instance from 91.22 MB:
https://hub.docker.com/layers/xanimo/1.14.5-dogecoin/bullseye/images/sha256-aa8fc7d558018725d2587daf45156db1412e50eec65bf23711274f2a4d172136?context=explore

to 57.11 MB without removing WORKDIR /tmp and COPY ...tar.gz ./:
https://hub.docker.com/layers/xanimo/1.14.5-dogecoin/no-qt/images/sha256-e6256b9c177e45dfdcd191cb332a1eabce931b475ff6104e8ba4d28880dbb5b9?context=explore

to 47.62 MB in it's current state as reflected in this PR:
https://hub.docker.com/layers/xanimo/1.14.5-dogecoin/headless/images/sha256-fde519015d6bce653dd7619c3f543be9bd4665a36d4174ddefb32acde512d99a?context=explore

.gitignore
```diff
@@ -0,0 +1,4 @@
+*.py[cod]
+*/*/*/__pycache__/
+*/*/__pycache__/
+*/__pycache__/
```
AbcSxyZ: It's enough to go for it
```suggestion
__pycache__/
```

1.14.5/bullseye/Dockerfile
```diff
@@ -53,7 +53,11 @@ RUN RLS_FILE_NAME=$(cat .filename) \
        | grep OK | shuf -n 1 | sed s/:.*// > random_signer.txt \
     && grep ${RLS_FILE_NAME} sigs/${SIG_PATH}/$(cat random_signer.txt)/*assert | sha256sum -c \
     && grep ${RLS_FILE_NAME} SHASUMS | sha256sum -c \
-    && mv ${RLS_FILE_NAME} dogecoin.tar.gz
+    && tar -xzf ${RLS_FILE_NAME} \
+    --wildcards "*/dogecoin*" \
+    --exclude="*-qt*" \
+    --strip-components=1 \
+    --one-top-level=dogecoin
```
AbcSxyZ: As a reminder, we would be able to simplify tar extraction with https://github.com/dogecoin/docker/pull/34, when we won't need man pages anymore.

1.14.5/bullseye/Dockerfile
```diff
@@ -67,27 +71,17 @@ RUN useradd ${USER} --home-dir ${HOME}
...
-    && chmod 4555 /usr/local/bin/dogecoin* \
-    && rm -rf *
+COPY --from=verify /verify/dogecoin/share/man/man1/*.1 /usr/share/man/man1/
 
-COPY entrypoint.py /usr/local/bin/entrypoint.py
-RUN chmod 500 /usr/local/bin/entrypoint.py
+COPY --from=verify --chown=${USER}:${USER} --chmod=4555 /verify/dogecoin/bin/dogecoin* /usr/local/bin/
 
-WORKDIR ${HOME}
+COPY --chmod=500 entrypoint.py /usr/local/bin/entrypoint.py
```
AbcSxyZ: We stay with a chmod in 2 steps and not in COPY because `COPY --chmod` is a buildx feature, it would disable the use of `docker build`.

1.14.5/bullseye/entrypoint.py
```diff
@@ -121,7 +120,7 @@ def run_executable(executable, executable_args):
     to manage a single process in a container & more predictive
     signal handling.
     """
-    if executable in ["dogecoind", "dogecoin-qt"]:
+    if executable in ["dogecoind"]:
```
AbcSxyZ:
```suggestion
    if executable == "dogecoind":
```

AbcSxyZ (changes requested):
A good opportunity to speak about what we want to do regarding `dogecoin-qt`.

It should be possible to containerize graphical application (did you tried it ?), by potentially sharing the window system of the host. I saw [some articles related to it](https://www.google.com/search?q=docker+graphical+application), I tried briefly few months ago but failed, I didn't go really far.

But it will probably not be the main use and be really a really specific feature. I didn't see the `dogecoin-qt` size, which is ~30MB on amd64. As you show, removing this executable divide almost the size of the image by 2, which is really significant.

Idk if adding GUI worth it for a Dogecoin Core image, if we could get rid of it for now and eventually if there is some demand in the future think about adding it ? Maybe it could be also a `qt` tag/image, to be able to provide `dogecoin-qt` through docker with keeping the benefit of a small size image only with other executables as main image.

patricklodder (changes requested):
Need some changes:
1. Please put changes to `*ignore` files in a separate pull request, has nothing to do with removing QT
2. Please further separate this into 2 proposals:
    1. removal of `dogecoin-qt` - this is easy to approve
    2. changing the way we copy the files into the final container - this needs much more discussion and I have a lot of comments on that particular code.

patricklodder:
Re: whether we want QT.

There is no dogecoin-qt for ARM architectures for dogecoin 1.14.5. So we can only provide it on a subset of images right now, I'm not really a fan.

> It should be possible to containerize graphical application (did you tried it ?), by potentially sharing the window system of the host. I saw some articles related to it, I tried briefly few months ago but failed, I didn't go really far.

I'd expect that we'd need to include some X11 libraries in the docker image for that to work and that significantly increases attack/maintenance surface. I'd recommend that if there is a demand for this, to build it in a separate variant rather than in the main image? As a node operator, I would not want my dogecoind deployment to be bloated with X11 stuff.

1.14.5/bullseye/entrypoint.py
xanimo: done

xanimo:
> Need some changes:
> 
>     1. Please put changes to `*ignore` files in a separate pull request, has nothing to do with removing QT
> 
>     2. Please further separate this into 2 proposals:
>        
>        1. removal of `dogecoin-qt` - this is easy to approve
>        2. changing the way we copy the files into the final container - this needs much more discussion and I have a lot of comments on that particular code.

done

xanimo:
> Re: whether we want QT.
> 
> There is no dogecoin-qt for ARM architectures for dogecoin 1.14.5. So we can only provide it on a subset of images right now, I'm not really a fan.
> 
> > It should be possible to containerize graphical application (did you tried it ?), by potentially sharing the window system of the host. I saw some articles related to it, I tried briefly few months ago but failed, I didn't go really far.
> 
> I'd expect that we'd need to include some X11 libraries in the docker image for that to work and that significantly increases attack/maintenance surface. I'd recommend that if there is a demand for this, to build it in a separate variant rather than in the main image? As a node operator, I would not want my dogecoind deployment to be bloated with X11 stuff.

i personally prefer headless operation over gui so i share your sentiment. i think we should potentially support this in the future but at the moment i'm not interested in maintaining or developing such an image.

1.14.5/bullseye/Dockerfile
```diff
@@ -77,7 +77,7 @@ COPY --from=verify /verify/dogecoin.tar.gz ./
 
 # Move downloaded binaries and man pages in the container system.
 # Setuid on binaries with $USER rights, to limit root usage.
-RUN tar -xvf dogecoin.tar.gz --strip-components=1 \
+RUN tar -xvf dogecoin.tar.gz --exclude="*-qt*" --strip-components=1 \
```
patricklodder: Recommend to not do this, and instead explicitly copy the files we need, as that is less error-prone. On line 82, can do `cp bin/dogecoind bin/dogecoin-cli bin/dogecoin-tx /usr/local/bin/` instead of wildcard? Same on line 81...

patricklodder (changes requested):

xanimo:
> A good opportunity to speak about what we want to do regarding `dogecoin-qt`.
> 
> It should be possible to containerize graphical application (did you tried it ?), by potentially sharing the window system of the host. I saw [some articles related to it](https://www.google.com/search?q=docker+graphical+application), I tried briefly few months ago but failed, I didn't go really far.
> 
> But it will probably not be the main use and be really a really specific feature. I didn't see the `dogecoin-qt` size, which is ~30MB on amd64. As you show, removing this executable divide almost the size of the image by 2, which is really significant.
> 
> Idk if adding GUI worth it for a Dogecoin Core image, if we could get rid of it for now and eventually if there is some demand in the future think about adding it ? Maybe it could be also a `qt` tag/image, to be able to provide `dogecoin-qt` through docker with keeping the benefit of a small size image only with other executables as main image.

i can't re-request a review from you so pinging you here. my thoughts on this are here: https://github.com/dogecoin/docker/pull/35#issuecomment-991796527

1.14.5/bullseye/Dockerfile
xanimo: done

patricklodder:
Blocked by #34 as that removes manpages but adds some `dogecoin-qt`

xanimo:
> Blocked by #34 as that removes manpages but adds some `dogecoin-qt`

removed copying manpages and ref of dogecoin-qt in entrypoint.

patricklodder (approved):
tested ACK.

Full x86_64 image went from 195MB at dc0a034 to 160MB with this PR

patricklodder:
@AbcSxyZ could you please let us know if you:

1. Agree with the concept/direction of this change (much like you proposed yourself in your last review comment, this PR implies that a dogecoin-qt deployment must be a separate image)
2. Agree with the implementation done here?

Thanks 🙏

AbcSxyZ (approved):
ACK for me, a good thing to remove `dogecoin-qt`.

PR #38, opened 2021-12-11
feat/ignore: adds *ignore files

xanimo:
.dockerignore to omit unnecessary and potentially sensitive files from being added into our image.
.gitignore to omit pycache and generated python test files

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.github
+.gitignore
+.cache
+*.py[cod]
+__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.cache
+*.py[cod]
+__pycache__/


.dockerignore
```diff
@@ -0,0 +1,7 @@
+.git
+.github
+.gitignore
+.cache
```
patricklodder, line 4: this probably needs to go into `.gitignore` too?

.dockerignore
```diff
@@ -0,0 +1,7 @@
...
+1.14.5/
+tests/
+tools/
```
patricklodder: why?

patricklodder: please add newline

.gitignore
```diff
@@ -0,0 +1,2 @@
+*.py[cod]
+__pycache__/
```
patricklodder: ignore also in `.dockerignore`?

patricklodder: please add newline

patricklodder (changes requested):

.dockerignore
xanimo: done

.dockerignore
xanimo: removed. disregard for now, i need to do more research.

xanimo: hmm i added a newline on both and for some reason they're not reflecting on either?

.gitignore
xanimo: done

patricklodder (approved):

AbcSxyZ (approved):

PR #39, opened 2021-12-11
refactor: mod cp tar.gz using buildx for leaner builds

xanimo:
refined Dockerfile by extracting the specific files we need from our archive and copying each while setting appropriate permissions and owner:group in each COPY.

amd64 63.36 MB
https://hub.docker.com/layers/182012868/xanimo/1.14.5-dogecoin/buildx-lean/images/sha256-02d46eed6bf4e35a21c7873b938a5be46d5526e895b21423a2d758213d24da89?context=repo

compared to current 91.02 MB
https://hub.docker.com/layers/180643304/xanimo/1.14.5-dogecoin/bullseye/images/sha256-df482148bac8ebe4861898b357d609533b4847e514d01bc76a7cdb2479c071ca?context=repo


patricklodder:
requires buildkit which is not shipped everywhere by default. Cannot do the `--chmod` part of this. See https://github.com/dogecoin/docker/pull/14#discussion_r758378325

patricklodder:
Blocked by #35

Issue #21, opened 2021-12-01
apt add `--no-install-recommends` flags


AbcSxyZ:
Reduce downloaded/installed size of `apt` packages by adding `--no-install-recommends`.

patricklodder:
On its own, this saves 2MB in the final image on my test runs. I'll PR this, but took a quick look and we can optimize elsewhere too. I'll create a separate issue for that.

AbcSxyZ:
Fixed with https://github.com/dogecoin/docker/pull/41

PR #41, opened 2021-12-12
deps/no-install-recommends

xanimo:
fix [DL3015](https://github.com/hadolint/hadolint/wiki/DL3015). gpg-agent is not installed when using --no-install-recommends so we install it with gpg.


AbcSxyZ (approved):
ACK, fix https://github.com/dogecoin/docker/issues/21

patricklodder (approved):
tested ACK

PR #42, opened 2021-12-12
deps/apt-get

xanimo:
fix [DL3027](https://github.com/hadolint/hadolint/wiki/DL3027)

PR #44, opened 2021-12-12
Fix recommendations from hadolint

patricklodder:
Fixes:

- find -> xargs vulnerability
- separate wildcard args for rm
- non-quoted (parts of) filenames
  - even though we can be pretty sure that there won't be any spaces or unicode characters in filenames we generate, we cannot be 100% sure, so it's better to just protect everything. 
- make wget use a progress bar
  - I would prefer `curl` because it's cleaner but that's just a personal preference so I ignored myself for now. May need to look at this again from a security perspective
- fail inside pipes
  - Note that `gverify` can fail so I explicitly negate that, but already introduced a check with the filename quote fix, so this is checked procedurally instead.

I left the `apt` related issues open because those rely on #42.

### DIFF
--- a/1.14.5/bullseye/Dockerfile
+++ b/1.14.5/bullseye/Dockerfile
@@ -48,7 +48,7 @@ RUN set -ex && ARCHITECTURE=$(dpkg --print-architecture) \
     && if [ "${ARCHITECTURE}" = "i386" ]; then RLS_ARCH=i686-pc; fi \
     && if [ "${RLS_ARCH}" = "" ]; then echo "Could not determine architecture" >&2; exit 1; fi \
     && RLS_FILE_NAME="dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz" \
-    && wget "${RLS_LOCATION}/${RLS_FILE_NAME}" \
+    && wget -q --show-progress --progress=bar:force:noscroll "${RLS_LOCATION}/${RLS_FILE_NAME}" \
     && GITIAN_OUTPUT=$(gitian/bin/gverify --no-markup -d sigs -r "${SIG_PATH}" "${DESCRIPTOR_PATH}") \
     && RANDOM_SIGNER=$(echo "${GITIAN_OUTPUT}" | grep OK | shuf -n 1 | sed s/:.*//) \
     && if [ "${RANDOM_SIGNER}" = "" ]; then echo "No valid signers found" >&2; exit 1; fi \
--- a/1.14.5/bullseye/Dockerfile
+++ b/1.14.5/bullseye/Dockerfile
@@ -47,13 +47,15 @@ RUN set -ex && ARCHITECTURE=$(dpkg --print-architecture) \
     && if [ "${ARCHITECTURE}" = "armhf" ]; then RLS_ARCH=arm && RLS_LIB=gnueabihf; fi \
     && if [ "${ARCHITECTURE}" = "i386" ]; then RLS_ARCH=i686-pc; fi \
     && if [ "${RLS_ARCH}" = "" ]; then echo "Could not determine architecture" >&2; exit 1; fi \
-    && RLS_FILE_NAME=dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz \
-    && wget ${RLS_LOCATION}/${RLS_FILE_NAME} \
-    && gitian/bin/gverify --no-markup -d sigs -r ${SIG_PATH} ${DESCRIPTOR_PATH} \
-       | grep OK | shuf -n 1 | sed s/:.*// > random_signer.txt \
-    && grep ${RLS_FILE_NAME} sigs/${SIG_PATH}/$(cat random_signer.txt)/*assert | sha256sum -c \
-    && grep ${RLS_FILE_NAME} SHASUMS | sha256sum -c \
-    && mv ${RLS_FILE_NAME} dogecoin.tar.gz
+    && RLS_FILE_NAME="dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz" \
+    && wget "${RLS_LOCATION}/${RLS_FILE_NAME}" \
+    && GITIAN_OUTPUT=$(gitian/bin/gverify --no-markup -d sigs -r "${SIG_PATH}" "${DESCRIPTOR_PATH}") \
+    && RANDOM_SIGNER=$(echo "${GITIAN_OUTPUT}" | grep OK | shuf -n 1 | sed s/:.*//) \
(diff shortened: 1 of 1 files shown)


xanimo (dismissed):
nice you beat me to with regards to pipefail and everything with exception to separate wildcard args for rm and xargs vuln.

i'm totally fine switching from wget to curl if it saves an extraneous dep but will wait for your sec research and proposal to eval again.

just needs a rebase if merging #42 first, otherwise good to go! ACK!

1.14.5/bullseye/Dockerfile
```diff
@@ -33,11 +33,13 @@ RUN apt update && apt install --no-install-recommends -y \
     gpg-agent \
     && rm -rf /var/lib/apt/lists/*
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
AbcSxyZ: Maybe we could use globally, before the first `RUN` instruction:
```Dockerfile
SHELL ["/bin/bash", "-eux", "-o", "pipefail", "-c"]
```
You can see an explanation about [commands here](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425).

Briefly, to recap all:
-e: Exit whenever a command fails
-u: Use of unexisting variable fail
-x: Command printed
-o pipefail: A pipe with an invalid command inside will return 1 instead of 0.

It allows us to:
- remove `set` commands in RUN instructions
- Avoid variable check/error message/exit for `RLS_ARCH` & `RANDOM_SIGNER`

1.14.5/bullseye/Dockerfile
```diff
@@ -47,13 +49,15 @@ RUN set -ex && ARCHITECTURE=$(dpkg --print-architecture) \
...
     && if [ "${RLS_ARCH}" = "" ]; then echo "Could not determine architecture" >&2; exit 1; fi \
-    && RLS_FILE_NAME=dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz \
-    && wget ${RLS_LOCATION}/${RLS_FILE_NAME} \
-    && gitian/bin/gverify --no-markup -d sigs -r ${SIG_PATH} ${DESCRIPTOR_PATH} \
-       | grep OK | shuf -n 1 | sed s/:.*// > random_signer.txt \
-    && grep ${RLS_FILE_NAME} sigs/${SIG_PATH}/$(cat random_signer.txt)/*assert | sha256sum -c \
-    && grep ${RLS_FILE_NAME} SHASUMS | sha256sum -c \
-    && mv ${RLS_FILE_NAME} dogecoin.tar.gz
+    && RLS_FILE_NAME="dogecoin-${RLS_VERSION}-${RLS_ARCH}-${RLS_OS}-${RLS_LIB}.tar.gz" \
+    && wget -q --show-progress --progress=bar:force:noscroll "${RLS_LOCATION}/${RLS_FILE_NAME}" \
```
AbcSxyZ, line 54: I agree if a `curl -o <file> <url>` is good for the linter it's nicer.

We can also directly rename it to `dogecoin.tar.gz` during this line instead of the last `mv`.

AbcSxyZ (changes requested):
> Note that gverify can fail so I explicitly negate that, but already introduced a check with the filename quote fix, so this is checked procedurally instead.

You would like to enable the Dockerfile with no signer to avoid the issue we were speaking about, where only new signers would release the binary ? In this case it changes the use of `-u` variable in my comments, I suppose you didn't add it for this reason.

If this is the case, we may try to use a more explicit approach.

1.14.5/bullseye/Dockerfile
```diff
@@ -47,13 +49,15 @@ RUN set -ex && ARCHITECTURE=$(dpkg --print-architecture) \
...
+    && GITIAN_OUTPUT=$(gitian/bin/gverify --no-markup -d sigs -r "${SIG_PATH}" "${DESCRIPTOR_PATH}") || true \
```
AbcSxyZ, line 55: I see why you used `|| true`, gverify exit with 1 whenever a signature in the list is invalid, so probably always.

The script must have a `RANDOM_SIGNER`, it will stop without it. Related to my previous comment, we could use `set -u` and remove variable check/error message.
A bit annoying this gverify behavior.

1.14.5/bullseye/Dockerfile
patricklodder: Agree with removing `set -ex`, for now I'm not adding `-u` because I'm not convinced that "no variable checks" is better than what we have right now, we can look at that separately though. Will adjust.

1.14.5/bullseye/Dockerfile
patricklodder: > directly rename it to dogecoin.tar.gz

That would require a change in the logic for shasum checking that at this point I think will make things less readable. Let's consider curl separate from this PR, which is to comply with linters.

1.14.5/bullseye/Dockerfile
patricklodder: It's just that sigs got merged without checking, so there is a set of signatures that causes mismatch. This can happen also when there is contention, so it's good that it exits with an error. Just not entirely cool for our use case, but I don't think it should change for generic gverify. 

As an alternative, we can script a gitian verification process that suits our needs better, but with it nearing end-of-life, we may as well settle on negating the error exit for now.

xanimo: > It's just that sigs got merged without checking, so there is a set of signatures that causes mismatch. This can happen also when there is contention, so it's good that it exits with an error. Just not entirely cool for our use case, but I don't think it should change for generic gverify.
> 
> As an alternative, we can script a gitian verification process that suits our needs better, but with it nearing end-of-life, we may as well settle on negating the error exit for now.

i think this is fine but a note that fdov, @AbcSxyZ and my keys got merged after v1.14.5 tag so it's only pulling langerhans, michi, rnicoll, stapler117, whit and your keys:

that's why fdov, @AbcSxyZ and mine get:
```
gpg: Can'\''t check signature: No public key
xanimo: BAD SIGNATURE
```
because our key's aren't imported in `find dogecoin/contrib/gitian-keys -name '*.pgp' -exec gpg --import '{}' +` line. if we pulled from `-b 1.14.6-dev` we're included/can verify. regardless since nearing eol and having 4/6 gitian signers for v1.14.5 and only needing 1 to gverify against is fine and works to verify authenticity.

xanimo (approved):
ACK

xanimo:
@AbcSxyZ if you have a moment to review would be greatly appreciated :heart:

AbcSxyZ (approved):
utACK for me. Would be great to add some improvement afterward:
- Using curl & rename during the download
- `GITIAN_OUTPUT=$(gitian/bin/gverify --no-markup -d sigs -r "${SIG_PATH}" "${DESCRIPTOR_PATH}") || true` use true inside the command substitution for clarity imo
- I still think using `set -u` could simplify the Dockerfile, it could be great to discuss it.

Issue #50, opened 2021-12-17
Convert datadir pytest into CI tests


AbcSxyZ:
Convert test from #25 to CI tests using [integration test framework](https://github.com/dogecoin/docker/tree/main/tests/integration).
```python
def test_datadir(hook, host):
    """
    Verify if datadir can be changed and created proprely.

    Verify datadir metada, if it's owned by script user (dogecoin).
    """
    # Use a unique directory for this test
    tmp_datadir = "/tmp/datadir_test"
    datadir_argument = f"-datadir={tmp_datadir}"

    test_args = ["dogecoind", datadir_argument]
    test_environ = {}

    result_args = [
            pytest.abs_path("dogecoind"),
            datadir_argument,
            "-printtoconsole",
            ]
    result_environ = {}

    # Run test to generate datadir
    hook.test(test_args, test_environ, result_args, result_environ)
    assert hook.result == hook.reference

    # Test datadir metadata
    datadir_folder = host.file(tmp_datadir)
    assert datadir_folder.user == pytest.user
    assert datadir_folder.group == pytest.user
    assert datadir_folder.mode == 0o755
```

PR #51, opened 2021-12-17
DockerRunner: Enable background container to run commands against

AbcSxyZ:
Following a comment where @patricklodder was speaking about having a mechanism to start/stop a container for the entire test.

Implemented it in the idea to add some tests who can need it, at least datadir test from #50.

It changes the way commands are proceeded : A single background container is started for the entire test of a TestRunner subclass, and all commands for the test are executed with `docker exec` (calling the entrypoint directly).

## Run CI tests

Will need some documentation to facilitate usage, but for now it's possible to run it using:
```bash
python3 -m tests.integration_runner --platform amd64 --image [dogecoin-image] --version 1.14.5 [--verbose]
```

### DIFF
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -8,20 +8,28 @@ import subprocess
 import sys
 
 class DockerRunner:
-    """Run a docker container in background to execute testing commands"""
+    """
+    Interface with the Docker cli to build & launch `docker run` and
+    `docker exec` commands.
 
+    - `docker run` is used for a one shot container.
+    - `docker exec` enable a test to use the same background container for
+      multiples commands, launched automatically at first exec instruction.
+    """
     def __init__(self, platform, image, verbose):
         """Sets platform and image for all tests ran with this instance"""
         self.platform = platform
         self.image = image
         self.verbose = verbose
+        #Background container for `docker exec`
         self.container_id = None
-
-        # Launch the container
-        self._start()
 
     def execute(self, envs, args):
         """Launch `docker exec` commands inside the background container"""
+        # Create background container if not existing on first exec command
+        if self.container_id is None:
+            self._start_background()
+
         command = self._construct_command("exec", envs, args)
         return self._shell(command)
 
@@ -33,11 +41,10 @@ class DockerRunner:
         return self._shell(command)
 
     def __del__(self):
-        """Remove test container"""
-        stop_command = ["docker", "rm", "-f", self.container_id]
-        self._shell(stop_command, silent=True)
+        """Clean up background container if enabled"""
+        self._stop_background()
 
-    def _start(self):
+    def _start_background(self):
         """
         Launch a docker container. Done in 2 step using create + start.
 
@@ -57,6 +64,12 @@ class DockerRunner:
                 "docker", "start", self.container_id
                 ]
         self._shell(start_command, silent=True)
+
+    def _stop_background(self):
+        """Remove background test container if used"""
+        stop_command = ["docker", "rm", "-f", self.container_id]
+        if self.container_id:
+            self._shell(stop_command, silent=True)
 
     def _shell(self, command, silent=False):
         """Run an arbitrary shell command and return its output"""
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -80,7 +80,13 @@ class DockerRunner:
             command.append("-e")
             command.append(env)
 
-        command.append(self.container_id)
+        # Use container or image depending on command type
+        if docker_cmd == "exec":
+            tag = self.container_id
+        elif docker_cmd == "run":
+            tag = self.image
+
+        command.append(tag)
 
         # Launch all shell commands using entrypoint.py only
         command.append("entrypoint.py")
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -34,9 +34,7 @@ class DockerRunner:
         return self._shell(command)
 
     def run(self, envs, args):
-        """
-        Launch `docker run` commands to create a new container for each command
-        """
+        """Launch `docker run` commands, create a new container each time"""
         command = self._construct_command("run", envs, args)
         return self._shell(command)
 
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -50,7 +50,7 @@ class DockerRunner:
                 self.image,
                 "sleep", "infinity",
                 ]
-        self.container_id = self._shell(create_command, silent=True)
+        self.container_id = self._shell(create_command, silent=True)[:16]
 
     def _stop_background(self):
         """Remove background test container if used"""
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -45,25 +45,14 @@ class DockerRunner:
         self._stop_background()
 
     def _start_background(self):
-        """
-        Launch a docker container. Done in 2 step using create + start.
-
-        Keep the container running with a shell open thanks to `--interactive`,
-        having stdin open keep the process running.
-        """
+        """Start a background container to run `exec` commands inside"""
         create_command = [
-                "docker", "create",
+                "docker", "run", "-d",
                 "--platform", self.platform,
-                "--interactive",
                 self.image,
-                "/bin/bash",
+                "sleep", "infinity",
                 ]
         self.container_id = self._shell(create_command, silent=True)
-
-        start_command = [
-                "docker", "start", self.container_id
-                ]
-        self._shell(start_command, silent=True)
 
     def _stop_background(self):
         """Remove background test container if used"""
--- a/tests/integration/framework/docker_runner.py
+++ b/tests/integration/framework/docker_runner.py
@@ -8,49 +8,83 @@ import subprocess
 import sys
 
 class DockerRunner:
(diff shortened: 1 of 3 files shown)


patricklodder:
Is it possible to retain the single-shot run method so that we can separate framework updates from test updates?

AbcSxyZ:
Sure

AbcSxyZ:
I restored run commands. I used exec everywhere for now, where would you use run ? What kind of difference are you thinking of ?

patricklodder:
> What kind of difference are you thinking of ?

Currently, there is a difference between exec and run: run uses entry point which de-escalates the uid, exec does not have an entry point and we don't `USER` right now, so everything there is root. This yields different results so we need to retain both until we have a better security solution. 

Additionally, you're changing the version test right now, that's not needed with this enhancement.

AbcSxyZ:
Exec commands are also using entrypoint. Using verbose to show tests commands:
```
$ docker exec -e VERSION=1 c5a168d2a634dde84 entrypoint.py dogecoind
$ docker exec c5a168d2a634dde84 entrypoint.py dogecoin-cli -?
$ docker exec c5a168d2a634dde84 entrypoint.py dogecoin-tx -?
```

Actually, I also add the entrypoint manually when using `docker run`, it's done in _construct_command:
https://github.com/dogecoin/docker/blob/84b4f913cfeadd6ddce19d0d4ad14e1d380a2b97/tests/integration/framework/docker_runner.py#L75

**EDIT**: My runs commands sucks

With my fix, docker run launch the following commands:
```
$ docker run -e VERSION=1 dogecoin entrypoint.py dogecoind                                                                                                                      
$ docker run dogecoin entrypoint.py dogecoin-cli -?                                                                                                                             
$ docker run dogecoin entrypoint.py dogecoin-tx -?     
```

patricklodder:
> Exec commands are also using entrypoint. Using verbose to show tests commands:

That's not what we'd recommend a user to do though, right? I'll later on comment on that when I review the code.

AbcSxyZ:
Yes, it builds manually the command created by ENTRYPOINT + CMD.

tests/integration/framework/docker_runner.py
```diff
@@ -8,49 +8,93 @@
...
+    def run(self, envs, args):
+        """
+        Launch `docker run` commands to create a new container for each command
+        """
+        command = self._construct_command("run", envs, args)
+        return self._shell(command)
+
+    def __del__(self):
+        """Remove test container"""
+        stop_command = ["docker", "rm", "-f", self.container_id]
```
patricklodder: If we want to retain the container (to investigate because an error happened) it's useful to have a 2-step process: `docker stop` the container, and then `docker rm` the container. We can then choose to not rm on test failure, allowing a `docker start` and perhaps `exec` into a shell to investigate.

Can we split this up into 2 commands that are called independent from the destructor?

tests/integration/framework/docker_runner.py
```diff
@@ -8,49 +8,93 @@
...
+        self._shell(stop_command, silent=True)
+
+    def _start(self):
```
patricklodder: Why protected?

tests/integration/framework/docker_runner.py
```diff
@@ -8,49 +8,93 @@
...
         """
-        Construct a docker command with env and args
+        Launch a docker container. Done in 2 step using create + start.
+
+        Keep the container running with a shell open thanks to `--interactive`,
+        having stdin open keep the process running.
```
patricklodder: Why is this better than `docker run -d <params> sleep infinity`?

tests/integration/framework/docker_runner.py
```diff
@@ -8,49 +8,93 @@
...
-            command.append(arg)
-
-        return command
+        # Use container or image depending on command type
+        if docker_cmd == "exec":
+            tag = self.container_id
+        elif docker_cmd == "run":
+            tag = self.image
 
-    def run_interactive_command(self, envs, args):
```
patricklodder, line 36: could you please bring this method back so that the tests don't have to change?

tests/integration/framework/test_runner.py
```diff
@@ -48,6 +55,9 @@ def main(self):
         self.add_options(parser)
         self.options = parser.parse_args()
 
+        self.container = DockerRunner(self.options.platform,
+                self.options.image, self.options.verbose)
+
```
patricklodder: this forces all tests to run/exec a container. should leave it up to the individual tests in their `run_test` method, by putting this in a separate method that can be overridden by child classes and then call that method here.

tests/integration/framework/test_runner.py
```diff
@@ -25,15 +25,22 @@ def run_test(self):
         """Actual test, must be implemented by the final class"""
         raise NotImplementedError
 
-    def run_command(self, envs, args):
```
patricklodder, line 28: Please restore this so that version test doesn't have to change

tests/integration/version.py
```diff
@@ -27,21 +27,21 @@ def run_test(self):
         self.version_expr = re.compile(f".*{ self.options.version }.*")
 
         # check dogecoind with only env
-        dogecoind = self.run_command(["VERSION=1"], [])
```
patricklodder, line 30: Please revert all changes to this file.

patricklodder (changes requested):

tests/integration/version.py
AbcSxyZ: Using `self.docker_run` would be fine for you ?
Also, methods `TestRunner.docker_run` and `TestRunner.docker_exec` return directly the output to avoid the use of `.stdout` access & encoding/decoding stuff. Error messages are directly handled on command execution (DockerRunner._shell) and create test failure, not used in tests (at least for now). You would like to restore this also ?

tests/integration/framework/docker_runner.py
AbcSxyZ: It's a way to do that. I agree `sleep infinity` would simplify the process.

tests/integration/framework/docker_runner.py
AbcSxyZ: With our 2 type of command, `exec` & `run`:

- `run` commands outputs (stdout + stderr) are direclty handled. Stdout is returned, stderr displayed when error occurs (see `DockerRunner._shell`). Logs are already caught.
- `exec` commands do not add any logs, they seem to handle only main process outputs.

We can do this, but logs are already displayed, not sure where we would use `docker log`.

tests/integration/framework/docker_runner.py
AbcSxyZ: Just a habit because it wasn't used by outside. It was designed with the idea that DockerRunner will always start a container for a test. If the design change and we do not use it necessary as you suggest, it needs to change.

tests/integration/framework/test_runner.py
AbcSxyZ: Fine

tests/integration/framework/docker_runner.py
AbcSxyZ: Related to https://github.com/dogecoin/docker/pull/51#discussion_r771949065

tests/integration/framework/test_runner.py
AbcSxyZ: I changed it, the line is still (almost) the same, but it does not start automatically a background container.

It will start automatically on the first `docker exec` command, making background containers optional without the need to start the container manually when writing tests.

tests/integration/framework/docker_runner.py
patricklodder: I'm okay with it when it's designed to not be ran from outside ever because `<reason>`, but I do not really agree with the reasoning.

> It was designed with the idea that DockerRunner will always start a container for a test.

That's a design change of DockerRunner introduced with this PR, my comment is based on original design that is more permissive. The reason why I didn't design it like that with the initial build of the framework is because I can think of tests where we want to just `docker run` a single command and check the output, to check the exact, automated behavior of that (as was done with the version test), or, to run multiple docker containers to test an end-to-end use case, e.g. doing development on regtest, where there are scenarios where you need at least 2 nodes that connect to each other, for example when you want to work with getblocktemplate or auxpow RPC interfaces.

AbcSxyZ:
I did some fixes to reproduce real `docker run` commands. For versions tests with each command launched with both `run` and `exec`, it generates the following:
```
version
----------------------
$ docker run -e VERSION=1 --platform amd64 dogecoin
$ docker exec -e VERSION=1 acc889cd4db4eba2 entrypoint.py dogecoind
$ docker run --platform amd64 dogecoin dogecoin-cli -?
$ docker exec acc889cd4db4eba2 entrypoint.py dogecoin-cli -?
$ docker run --platform amd64 dogecoin dogecoin-tx -?
$ docker exec acc889cd4db4eba2 entrypoint.py dogecoin-tx -?
```

tests/integration/framework/test_runner.py
patricklodder: what if I want to change the way the container starts? i.e. test for permission escalation?

AbcSxyZ: Any example ? No idea how to perform this kind of test.

But you have the `docker run` interface as normally, and even if using entrypoint through `docker exec` any commands will run as root (outside of Dogecoin executables), as entrypoint is designed to accept arbitrary bash command.

georgeartem:
@patricklodder do the commits on 12.19.21 (b8d2171, 9b3ac80 etc.) satisfy the testing requirements?

PR #52, opened 2021-12-18
configure pylint for tests

AbcSxyZ:
Prepare pylint configuration for tests from #25. Linter is done in 2 times, once for the source code of images without any `pylintrc`, and then for tests which have more trouble for linters.

### DIFF
--- a/pylintrc.tests
+++ b/pylintrc.tests
@@ -1,0 +1,7 @@
+[entrypoint-tests]
+disable =   duplicate-code, # Tests having similar code
+            no-member,      # pytest global variables in conftest.py not detected
+            consider-using-with, # conftest temporary directory would need a with
+            import-error    # EntrypointHook unable to import entrypoint
+    
+    
--- a/pylintrc.tests
+++ b/pylintrc.tests
@@ -1,4 +1,4 @@
-[entrypoint-tests]
+[tests-configuration]
 disable =   duplicate-code, # Tests having similar code
             no-member,      # pytest global variables in conftest.py not detected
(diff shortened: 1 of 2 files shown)


pylintrc.tests
```diff
@@ -0,0 +1,7 @@
+[entrypoint-tests]
```
patricklodder: is it possible to specify this on the pytest module rather than the importing entry point test?

AbcSxyZ: I didn't rename this title from #25, I changed it to `tests-configuration`.

Unsure about what you're asking for, what  does "on the pytest module" mean and what do you mean by importing entrypoint test in this case ? You think about moving the pylintrc to the tests' location, doing something from pytest ?

Or configure the pylintrc per directory ? This is not possible for now, they are doing an entire development to enable configuration per directory (see [project roadmap](https://github.com/PyCQA/pylint/projects/1) and issue *"[Add different configuration for different subdirectories](https://github.com/PyCQA/pylint/issues/618)"*.

patricklodder: I'm looking for a way to ignore pytest and not our own code. Perhaps it's easier to not import pytest for linting and annotate the import with an ignore? I was mistakenly under the impression that the rc file was specifying these ignores for the entry point test module - due to the header.

patricklodder: For now I filed #53 to unblock this.

patricklodder (approved):
ACK, debt has to be managed through issue #53

xanimo (approved):
ACK

